PR #207, opened 2020-12-05
don't test sign driver in CI

stephtr:


isaak654:
@stephtr I'm noticing some CI errors in many commits recently, do you have any idea/suggestion?
[logs_184.zip](https://github.com/sandboxie-plus/Sandboxie/files/5884376/logs_184.zip)
[logs_188.zip](https://github.com/sandboxie-plus/Sandboxie/files/5884377/logs_188.zip)

stephtr:
The issue is that Qt (which is being used for Sandboxie Plus) can't be compiled in CI. I just submitted #469 in order to comment it out until a solution is found.

PR #260, opened 2020-12-22
German translations for strings of plus version

bastik-1001:
The two updated files contain the initial German translation of those strings. Should there be any issue with spacing anywhere in the GUI, I will have to shorten text a bit. The same goes for translations which do not fit the context.

All strings have been translated from scratch.

Some Alt key accelerators had to be changed in German due to the translation not containing such a character.


DavidXanatos:
I will get rid of the qsbieapi_??.ts and instead return only error codes and have the strings in the sandman component, thats more elegant.

PR #264, opened 2020-12-22
Minor templates adjustments

isaak654:
Minor corrections in some existing templates.


DavidXanatos:
I have manually merged your changes, but I cant merge the file like this you saved it as ASCI and sandboxie requirers the template file to be Unicode, see yourself the file after you edited it has lost half its size. 
Please update your copy of the file and when making further changes ensure you commit it with the required encoding.
You can use an editor like https://www.flos-freeware.ch/notepad2.html

isaak654:
> 
> 
> I have manually merged your changes, but I cant merge the file like this you saved it as ASCI and sandboxie requirers the template file to be Unicode, see yourself the file after you edited it has lost half its size.
> Please update your copy of the file and when making further changes ensure you commit it with the required encoding.
> You can use an editor like https://www.flos-freeware.ch/notepad2.html

I have a question for you, why you're using utf-16 instead of native utf-8? 

utf16le

I think this issue was caused by the UTF-16LE conversion performed automatically by github.

DavidXanatos:
Could be, the online editor is quite limited, I would recommend downloading the file editing it locally and than committing the change, you could use a tool like Tortois SVN

isaak654:
> 
> 
> Could be, the online editor is quite limited, I would recommend downloading the file editing it locally and than committing the change, you could use a tool like Tortois SVN

Ok, thanks. I've figured out how to bypass it.

DavidXanatos:
great 👍

isaak654:
Personally I still think there is some encoding problem in many txt files.
For example Sandboxie/msgs/Text-Danish-1030.txt has the same UTF-16LE encoding that I've found before in another file.
All the files stored in that folder have the same encoding, except the updated chinese file involved in issue #268.

NewKidOnTheBlock:
+1 Make it as easy as possible to send in contributions.

PR #275, opened 2020-12-24
Update German translation for plus version

bastik-1001:
Once that is merged, the German translation is ready to be used for the next release.


DavidXanatos:
I have updated the sources to the upcoming build 0.5.3 could you please update the translation, as some english strings have changed

PR #296, opened 2020-12-29
Lang de german 2020 12 29

bastik-1001:
This contains changes to include the new strings or those that changed.


DavidXanatos:
I have updated the sandman sources for the upcoming next build, could you please take a look and update the translation.

PR #318, opened 2021-01-04
Chinese Simplified File

yuhao2348732:
Chinese Simplified File


yuhao2348732:
also add CPanelView ：Copy Cell、Copy Row、Copy Panel，when right click the forms

DavidXanatos:
the next build with have a facility to localize those strings as well

DavidXanatos:
I have updated the sandman sources for the upcoming next build, could you please take a look and update the translation.

PR #327, opened 2021-01-06
Update German translation for plus version

bastik-1001:
Had fixed a typo before.

86d6ca6 is the one that contains all new strings and required changes. (Which is true for f724716 as well)


bastik-1001:
Is there a way for me to tell when "Dismiss" means basically "Close"? I translated all instances of "Dismiss" to be the same namely being dismissive about what is shown on the screen. For some it does not appear to be correct like "Program was started outside of the sandbox" (Forced programs disabled) in which it seems to be an information and nothing that the user can negate.

yuhao2348732:
dismiss means ignore

DavidXanatos:

bastik-1001:
@DavidXanatos I used your forum to send you the required information.

DavidXanatos:
@bastik-1001 i did not get it, just send it on my gmail address

DavidXanatos:
found it seam to have ended up in the spam filter, you should get a invite from slack shortly

PR #328, opened 2021-01-07
Updated Russian translation

lufog:


DavidXanatos:

lufog:
I'm sorry, but I do not use this service. In addition, I cannot always answer on time, as sometimes I am busy with work.

PR #345, opened 2021-01-08
Update sandman_pt.ts

JNylson:
updated strings without translation


DavidXanatos:

PR #361, opened 2021-01-11
Complete another string

bastik-1001:
I missed that there was a two-part translation, but nothing to worry about.


DavidXanatos:

PR #363, opened 2021-01-11
Fixed a ton of German spelling errors in the translation

NewKidOnTheBlock:
I have a few questions, though:
What does `<oldsource>` mean? Why is that still in there? There are a ton of spelling errors contained in these tags which I didn't touch. E.g. in
[sandman_de.ts#L55](https://github.com/sandboxie-plus/Sandboxie/blob/b1980e4c3f831085a711242a8b57df4081582eff/SandboxiePlus/SandMan/sandman_de.ts#L55)
[sandman_de.ts #L307](https://github.com/sandboxie-plus/Sandboxie/blob/688034f482bef7795563e7907b517196787f1d31/SandboxiePlus/SandMan/sandman_de.ts#L307)
[sandman_de.ts #L391](https://github.com/sandboxie-plus/Sandboxie/blob/688034f482bef7795563e7907b517196787f1d31/SandboxiePlus/SandMan/sandman_de.ts#L391)
etc. Basically every single occurrence of `<oldsource>` has some kind of orthographical error in it.

I've also fixed a couple English spelling errors that came in `<source>` tags.
-> If I change the spelling of something in `<source>` tags, does the translation still work?

### DIFF
--- a/SandboxiePlus/SandMan/sandman_de.ts
+++ b/SandboxiePlus/SandMan/sandman_de.ts
@@ -202,15 +202,15 @@
     <message>
         <location filename="Windows/OptionsWindow.cpp" line="1894"/>
         <source>Please enter an auto exec command</source>
-        <translation>Bitte geben Sie einen Autoausführen-Kommndo ein</translation>
+        <translation>Bitte geben Sie einen Autoausführen-Kommando ein</translation>
     </message>
     <message>
         <location filename="Windows/OptionsWindow.cpp" line="2095"/>
         <source>This template is enabled globally. To configure it, use the global options.</source>
-        <translation>Diese Vorlage ist global aktiv, um sie zu konfigueren müssen die globalen Optionen genutzt werden.</translation>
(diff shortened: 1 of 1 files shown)


bastik-1001:
Thanks a bunch, it is quite embarrassing how many spelling mistakes I manged to include. In retrospect I should not have tried to force myself to rush to provide the translation. For some elements I tried to stay too close to the source. The changes are much better to understand in these instances. You should add NewKidOnTheBlock as translator in the readme, he deserves credit for that.

Just one thing the German word "Stopp" is indeed written with "pp", every German stop sign spells it "Stop", even though its description spells it with two p.

In line 2525/2526 the word "Vorlange" is spelled wrong. Should be "Vorlagen", but nothing compared to my incredible amount of typos.

DavidXanatos:

NewKidOnTheBlock:
@DavidXanatos
a) I've scanned through the _de file and found 31 occurrences of `<translation type="unfinished">`
I assume these got priority?

b) Is there also a _de file for SB Classic?


c) I still don't know what function the `oldsource` lines serve. Examples:
old1
old2
To me, these look redundant and are usually filled with errors.
Could we simply delete them and be done with it?

PR #402, opened 2021-01-18
Fixed German spelling for Text-German-1031.txt

NewKidOnTheBlock:
There was a warning message though:
"We’ve detected the file encoding as UTF-16LE. When you commit changes we will transcode it to UTF-8."

PR #452, opened 2021-01-27
UTF-8 BOM encoding

isaak654:
This should fix a corruption issue in some text files with UTF-16 encoding and mentioned in https://github.com/sandboxie-plus/Sandboxie/pull/402


isaak654:
I've just noticed that Sandboxie rejects templates.ini with UTF-8 encoding, but not UTF-8 with BOM.
I'm going to revert the changes to push another commit with the proper conversion.

PR #469, opened 2021-01-30
Disable Sandboxie-Plus build

stephtr:
As long as Qt doesn't compile in CI, let's disable it.


isaak654:
Please evaluate #471 before merging this.

PR #484, opened 2021-01-31
Update Templates.ini & multiple strings

isaak654:
This commit removes duplicated Seamonkey entries and some separators no longer needed in Templates.ini + various fixes.


DavidXanatos:
@isaak65 you'd you test is there are any issues using a utf8 encoded templates file as the driver supports that now, it would be nice to see the changes in github

isaak654:
> 
> 
> you'd you test is there are any issues using a utf8 encoded templates file as the driver supports that now, it would be nice to see the changes in github

I have no problem with UTF8 and UTF8-BOM... I don't know which one is better between them, but they are both fine.

Issue #526, opened 2021-02-04
File Copy Limit wraps at 4GB


typpos:
From #517.

v0.6.7 
v5.47.1

Sandboxie wraps File Copy Limit at 4GB. Affects Classic & Plus for releases from 31 Jan 2021.

Repro:
* Enter 4194303 KB for a sandbox, then open/save a 10KB file. OK
* Enter 4194304 KB for a sandbox, then open/save a 10KB file. Fail

PR #527, opened 2021-02-04
(#526) Fix: CopyLimit overflows at 4GB.

typpos:
Fix File Copy overflow/wrap.

### DIFF
--- a/Sandboxie/core/dll/file_copy.c
+++ b/Sandboxie/core/dll/file_copy.c
@@ -125,7 +125,7 @@ found_match:
         return mode;
 
     //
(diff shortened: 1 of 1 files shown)


Sandboxie/core/dll/file_copy.c
```diff
@@ -128,7 +128,7 @@ _FX ULONG File_MigrateFile_GetMode(const WCHAR* TruePath, ULONGLONG file_size)
     // if tere is no configuration for this file type/path decide based on the file size
```
isaak654, line 128:
```suggestion
    // if there is no configuration for this file type/path, then decide based on the file size
```

PR #534, opened 2021-02-06
Update Templates.ini

### LINKED ISSUE #499: Lingerer update_notifier.exe
From Vivaldi browser

stdedos:
* Add Google Chrome's `software_reporter_tool.exe` telemetry
* Add Vivaldi's `update_notifier.exe`
* Add `ssh-agent.exe`

* Remove trailing whitespace.

Fixes #499, and then some.

Signed-off-by: Stavros Ntentos <133706+stdedos@users.noreply.github.com>


stdedos:
@DavidXanatos A small token of changes (for now, I hope)

On this matter:
* Does it make sense / is it easy to add blocked programs? (`software_reporter_tool.exe`)
* "Should we somehow" fix those UTF-16 files? (https://git-scm.com/docs/gitattributes/2.19.2#_working_tree_encoding)
  `git diff` is complicated otherwise

isaak654:
@stdedos 
If you agree, I suggest a possible fix regarding [this](https://www.wilderssecurity.com/threads/released-sandboxie-plus-sbie-fork-versions-with-signed-driver.434924/page-35#post-2988714) issue related to Opera browser bookmarks.

Find:
https://github.com/sandboxie-plus/Sandboxie/blob/47d11ca78387419b1f14e0c27fe898efc816ace1/Sandboxie/install/Templates.ini#L82

Replace with:
`Tmpl.Opera=%AppData%\Opera Software\Opera Stable`

stdedos:
> If you agree, I would suggest to include (after some testing) a possible fix here regarding [this issue](https://www.wilderssecurity.com/threads/released-sandboxie-plus-sbie-fork-versions-with-signed-driver.434924/page-35#post-2988714).

That makes it complicates for me, since I cannot test it (I don't have the browser), neither am I willing to do so 😛

I am dragging my feet with a "very specific ticket", I don't see myself taking on more responsibilities 😕

DavidXanatos:
> 
> 
> @DavidXanatos A small token of changes (for now, I hope)
> 
> On this matter:
> 
>     * Does it make sense / is it easy to add blocked programs? (`software_reporter_tool.exe`)
> 
>     * "Should we somehow" fix those UTF-16 files? (https://git-scm.com/docs/gitattributes/2.19.2#_working_tree_encoding)
>       `git diff` is complicated otherwise

I'm not of any radar 😛

You can add blocked programs like this
```ClosedIpcPath=software_reporter_tool.exe,*```

the uft files are already fixed kind of, that is since build 0.6.5 the driver understands (hopefully flawless) UTF8 with BOM encoded ini files.
Just that in case it does have some unexpected issues the files were left for now in the old UTF16 formatting.

Sooo... feal free to save them as UTF8 with BOM and thoroughly test if it really works fine I think if enough people test we could release a utf8 encoded templates.ini with the next build

**Comment by stdedos** [comment]
> Sooo... feal free to save them as UTF8 with BOM and thoroughly test if it really works fine I think if enough people test we could release a utf8 encoded templates.ini with the next build

"Thoroughly test" is a complicated term 😛

I can push a PR, and then I will convince myself to install the CI build (are they accessible?) and use it as normal (until you post a "significantly better newer" version)

**Comment by DavidXanatos** [comment]
I think thay are not but I don't know why, or don't every fork of the repo has own CI with for them accessible build?

anyhow you don't need a new build just stop the driver, save the Templates.ini and Sandboxie.ini as utf 8 ans start it again

**Comment by stdedos** [comment]
> ... or don't every fork of the repo has own CI with for them accessible build?

I don't know how it works, but I have to say "maybe not".
As a random example, repo secrets don't get copied with forking 😃 

> anyhow you don't need a new build just stop the driver, save the Templates.ini and Sandboxie.ini as utf 8 ans start it again

Cool! 😀 Doing it now then

**Comment by isaak654** [comment]
> 
> That makes it complicates for me, since I cannot test it (I don't have the browser), neither am I willing to do so 😛
> 
> I am dragging my feet with a "very specific ticket", I don't see myself taking on more responsibilities 😕
> 
> Furthermore, I don't see the difference:
> 
> ```
> Tmpl.Opera=%AppData%\Opera Software\*\*
> Tmpl.Opera=%AppData%\Opera Software\Opera Stable
> ```

@stdedos
Well, I don't see the difference either. I've already tested these two paths separately, and my opera bookmarks are still saved and kept in both cases, so it doesn't cost me anything to change the first path with the second path if someone could benefit from it.

I'll create another pull request about it.

**Comment by stdedos** [comment]
> Well, I don't see the difference either. I've already tested these two paths separately, and my opera bookmarks are still saved and kept in both cases, ... 

If both cases work, then what's the point of the change? 😕 

Lacking testing, only if there is strong indication of necessity (eg Opera decided that they need to move their path, users come with pitchforks) would I have made the change myself. 

It sounds more safe to add a second Opera instance, or, if this would start being a pattern, array-expansion that would roughly do (`','.join(array_of_expansions)`) 

> ... so it doesn't cost me anything to change the first path with the second path if someone could benefit from it.
 
Change doesn't cost anything to develop; it I am afraid, however, whether it would cost something to an end user or not. It's also a silent change too; so a user might not notice it until too late. 

> I'll create another pull request about it.

I am sorry. I know how frustrating and time-consuming it must be, especially compared to the size of the change. The process that I follow is basically, if I commit changes only when I can assert them med-to-high level, and that does not sound like it.

**Comment by isaak654** [comment]
@stdedos 
You're right, probably the safest option is to add another OpenFilePath in the opera bookmarks template.

I'm evaluating some of the changes you made in this file (considering the annoying UTF-16 encoding that leads to longer times) and I don't agree with the creation of this new template:

```
[Template_Vivaldi_Updater]
Tmpl.Title=#4338,Vivaldi
Tmpl.Class=WebBrowser
LingerProcess=update_notifier.exe
```

1) `Tmpl.Title=#4338,Vivaldi` is basically the same title taken from the previous template, end users could be confused by this duplication.
2) Why creating a template for just one LingerProcess that could be added in the general `Template_LingerPrograms` list (that is automatically applied at Sandboxie's startup)? Most users just run browsers without ever seeing browsers templates.

stdedos:
> 1. `Tmpl.Title=#4338,Vivaldi` is basically the same exact title taken from the previous template, end users could be confused by this duplication.

Oh, fuck 😞 I missed the fact that all of those templates had a different `Tmpl.Title=#0000,vivaldi.exe`.

> 2. Why creating a template for just one LingerProcess that could be added in the general `Template_LingerPrograms` list (that is automatically applied by default)?

I thought that there are application groups (`vivaldi.exe`), which may inherit from another one, and they are "all applied together", kinda like how "Software Compatibility" looks like:

image

All of that, was done in the name of "why bulk all the programs in one template, when you can add them in a group that will give them more metadata (`update_notifier.exe` belongs to `vivaldi.exe`)?"

isaak654:
Good news, I've just managed to reproduce the opera bookmarks issue by adding some bookmarks outside the sandbox and at least one bookmark inside the sandbox.

PR #549, opened 2021-02-12
Fix: elevated-rights-handle visible in user space.

typpos:
TargetHandle references an address in user space and transiently contains a handle with elevated rights. A user thread can steal the handle by constantly polling with InterlockedCompareExchange.

Please review carefully. I'm not set up to test-run driver builds.


DavidXanatos:
Thanks I'll check it out asap

PR #564, opened 2021-02-15
Fix byte-len vs wchar-len error

typpos:
log_len is initially treated like len in wchars, later from lines 1037++ it's treated like len in bytes.

For example, if args->log_len.val == 2, you get a wrap-around in 1037, an out-of-bounds and out-of-ProbeForWrite write error in 1040.

### DIFF
--- a/Sandboxie/apps/control/BoxPage.cpp
+++ b/Sandboxie/apps/control/BoxPage.cpp
@@ -966,7 +966,7 @@ void CBoxPage::Appearance_OnOK(CBox &box)
         BOOL enable = (pCheckBox3->GetCheck() == BST_CHECKED ? TRUE : FALSE);
         BOOL title  = (pCheckBox4->GetCheck() == BST_CHECKED ? TRUE : FALSE);
         CString str;
-        GetDlgItem(ID_MIGRATE_KB)->GetWindowText(str);
+        GetDlgItem(ID_BORDER_WIDTH)->GetWindowText(str);
         int width = _wtoi(str);
         ok = box.SetBorder(enable, Appearance_BorderColor, title, width);
     }
--- a/Sandboxie/core/drv/session.c
+++ b/Sandboxie/core/drv/session.c
@@ -974,11 +974,11 @@ _FX NTSTATUS Session_Api_MonitorGetEx(PROCESS *proc, ULONG64 *parms)
     if (log_tid != NULL)
         ProbeForWrite(log_tid, sizeof(ULONG64), sizeof(ULONG64));
 
-	log_len = args->log_len.val / sizeof(WCHAR);
+	log_len = args->log_len.val / sizeof(WCHAR) * sizeof(WCHAR);
     if (!log_len)
         return STATUS_INVALID_PARAMETER;
 	log_data = args->log_ptr.val;
-    ProbeForWrite(log_data, log_len * sizeof(WCHAR), sizeof(WCHAR));
+    ProbeForWrite(log_data, log_len, sizeof(WCHAR));
 
     *log_type = 0;
 	if (log_pid != NULL)
--- a/Sandboxie/core/drv/session.c
+++ b/Sandboxie/core/drv/session.c
@@ -504,7 +504,8 @@ _FX NTSTATUS Session_Api_DisableForce(PROCESS *proc, ULONG64 *parms)
     in_flag = args->set_flag.val;
     if (in_flag) {
         ProbeForRead(in_flag, sizeof(ULONG), sizeof(ULONG));
-        if (*in_flag) {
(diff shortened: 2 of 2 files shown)


typpos:
Second commit. Unrelated to first:

Remove double-fetch of in_flag from user space.

typpos:
General related question:

The user-to-kernel data transfer pattern is something like..

```
ProbeForRead(userdata, ...)
kerneldataAB = userdata;
do stuff with kerneldataA.
if(kernelDataA...) do stuff with kerneldataB.
```

What ensures that the compiler does not re-order this or completely skips kerneldata and repeatedly refers to userdata? There is no "volatile userdata", no memory or instruction-reorder barriers, etc. I looked at the assembly output of a number of methods and they all look fine (subject to Intel x86 instruction scheduling), but I can't find anything in the C language specification that guarantees this? Sorry if that's a silly question, but my C isn't pro-grade.

DavidXanatos:
To the first issue:
```log_len = args->log_len.val / sizeof(WCHAR) * sizeof(WCHAR);``` 
can be replaced with 
```log_len = args->log_len.val``` 
I'll fix that in the next build for today its to late to re run everything.

To the second issue, its not always problematic to read the same user memory multiple times. ProbeForRead will throw an exception when it fails similarly to what happens when you later try to read the memory and it has became invalid. Its intended to provide a more defined behavior in case of everyday issues. 

In Api_FastIo_DEVICE_CONTROL you see that line 515 that calls the various functions is enclosed in a try block.
On an other case in Session_Api_MonitorGetEx we have an own try block as we always need to call Session_Unlock(irql); before exiting.

In Session_Api_DisableForce we don't do anything that would require clean up at the point we de-reference in_flag so making a copy is unnecessary.

About compiler reordering that shouldn't be an issue as the compiler shouldn't change the behavior of the code. 
About eliminating variables...
```
void func1(int* pvar1, int* pvar2)
{
func3();
int var1 = *pvar1;
func4();
int var2 = *pvar2;
func5();
if(var1)
func6();
etc....
}
```

One may think that the compiler could think to replace if(var1) with if(*pvar1) but if you have some function calls in between the compiler does not know what they do, they may tell some other portion of the program than the memory of pvar1 is not longer needed and can be freed. No compiler will to that extent investigate what which function does. So it should not think of doing that replacement.

```
void func1(int* pvar1, int* pvar2)
{
func3();
int var1 = *pvar1;
if(var1)
func6();
etc....
}
```

Now that's a different story, here it may indeed choose to optimize var1 away, but that this wouldn't be a problem eider.

So yet an other case

```
void func1(int* pvar1, int* pvar2)
{
func3();
int var1;
try{
var1 = *pvar1;
}
catch{var1=0}
if(var1)
func6();
etc....
}
```

here its clear that it cant go without a local copy as it may want to set it to 0;

so whats with that:
```
void func1(int* pvar1, int* pvar2)
{
func3();
int var1;
try{
var1 = *pvar1;
}
catch{
return;
}
if(var1)
func6();
etc....
}
```

Here the compiler should be intelligent enough to understand the purpose of the try block and also not try to remove the local copy.

Now this all does not mean there are no bugs in the optimization code, I remember 2010 or so VS had a few issues with compiling 64 bit code, resulting in re ordering that changed the behavior but these were bugs and are now fixed.


Generally the compiler shouldn't do any optimizations that change the behavior of the code, and it should not obtimize away variables if there is any chance they may be needed.

typpos:
Thanks, David, for spending so much time on this. Much appreciated.

Re item 1:

`
log_len = args->log_len.val / sizeof(WCHAR) * sizeof(WCHAR);
`

I didn't remove the divide & multiply because log_len must be even. If you trace your approach with `args->log_len.val == 1`, you will still get a wrap-around in 1038, and I didn't use `log_len = args->log_len.val & ~1`, because I couldn't sure sizeof(WCHAR) is always 2 and not 4.

Re item 2:

Understood. My concern was that double-fetch means an external thread can toggle the value and then 2 inconsistent decisions are made in the kernel. For example, if in_flag* == 0 in line 506, the ForceDisableAdmin check is skipped, but now in line 517 it may have been changed to DISABLE_JUST_THIS_PROCESS and it executes what normally would haven been blocked. Not a major issue here as much as I can tell but I don't know the full repercussions of this function.

Re item 3:

I'll have a closer look at your notes tomorrow, but I did run some compile-and-look-at-assembly trials with a simple sequence of variable assignments, like a=1, b=2, c=3 and the DO NOT always stay in order. Basically the compiler optimizes with the assumption of a single thread and does move things around, and it did that even when there was no obvious performance reason that I could see.

Thanks again
t

typpos:
New unrelated commit. Simple fix for sandboxie-classic crash on save-appearance.

PR #708, opened 2021-03-25
Add VLC Media Player and the MediaPlayer class to the templates

Dyras:
Hello! I've added VLC Media Player to the templates. I've created the new "MediaPlayer" class for it, and I'm planning on expanding that class with MPC-HC if you're OK with this pull request :)

Edit: These templates are waiting for this one to be submitted:
MPC-HC, MPC-BE, PotPlayer, KMPlayer, SMPlayer
SumatraPDF, Foxit Reader

I've also added an option for VLC to access the AppData in order to save the settings, and another one for access to the pictures folder to save screenshots.

Tested by me, and it works great.

### DIFF
--- a/Sandboxie/apps/control/AppPage.cpp
+++ b/Sandboxie/apps/control/AppPage.cpp
@@ -1020,13 +1020,6 @@ void CAppPage::AddPages(CPropertySheet &sheet, const CString &BoxName)
     info.WithLink = FALSE;
     info.WithCreate = FALSE;
     m_app_pages.AddTail(new CAppPage(&info, BoxName));
-    
-    info.ClassName = L"MediaPlayer";
-    info.TitleId = MSG_4391;
-    info.LabelId = MSG_4392;
-    info.WithLink = FALSE;
-    info.WithCreate = FALSE;
-    m_app_pages.AddTail(new CAppPage(&info, BoxName));
 
     info.ClassName = L"Print";
     info.TitleId = MSG_4224;
@@ -1046,6 +1039,13 @@ void CAppPage::AddPages(CPropertySheet &sheet, const CString &BoxName)
     info.TitleId = MSG_4228;
     info.LabelId = MSG_4207;
     info.WithLink = TRUE;
+    info.WithCreate = FALSE;
+    m_app_pages.AddTail(new CAppPage(&info, BoxName));
+    
+    info.ClassName = L"MediaPlayer";
+    info.TitleId = MSG_4393;
+    info.LabelId = MSG_4394;
+    info.WithLink = FALSE;
     info.WithCreate = FALSE;
     m_app_pages.AddTail(new CAppPage(&info, BoxName));
 
--- a/Sandboxie/install/Templates.ini
+++ b/Sandboxie/install/Templates.ini
@@ -2639,7 +2639,7 @@ Tmpl.Class=MediaPlayer
 OpenFilePath=vlc.exe,%AppData%\vlc\*
 
 [Template_VLC_DirectAccess_Photos]
-Tmpl.Title=#4395
+Tmpl.Title=#4395,VLC
 Tmpl.Class=MediaPlayer
 OpenFilePath=vlc.exe,%USERPROFILE%\Pictures\*
 
--- a/Sandboxie/install/Templates.ini
+++ b/Sandboxie/install/Templates.ini
@@ -2625,6 +2625,25 @@ OpenIpcPath=*\BaseNamedObjects*\ZoomTextRunning*
 OpenWinClass=ZT9MainWindow
 
 #
+# Media Players
+#
+
+[Template_VLC_Force]
+Tmpl.Title=#4395
+Tmpl.Class=MediaPlayer
+ForceProcess=vlc.exe
+
+[Template_VLC_DirectAccess_Profile]
+Tmpl.Title=#4396
+Tmpl.Class=MediaPlayer
+OpenFilePath=vlc.exe,%AppData%\vlc\*
+
+[Template_VLC_DirectAccess_Photos]
+Tmpl.Title=#4397
+Tmpl.Class=MediaPlayer
+OpenFilePath=vlc.exe,%USERPROFILE%\Pictures\*
+
+#
 # Download Managers
 #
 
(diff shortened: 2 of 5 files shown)


isaak654:
> 
> 
> I've created the new "MediaPlayer" class for it, and I'm planning on expanding that class with MPC-HC if you're OK with this pull request :)
> 

Good idea, but it's not enough to add a new class in the templates.ini. For example, I can't see those VLC templates represented in the UI of Sbie Classic.

I never did it before, but after browsing the source code, I think you need to apply the class in these files too:
https://github.com/sandboxie-plus/Sandboxie/search?q=EmailReader

isaak654 (changes requested):
It's not enough to add a new class in templates.ini. Only with these edits, VLC templates won't be represented in the UI of Sbie Classic.

After browsing the source code, I think you need to apply the new class in these files too: 
https://github.com/sandboxie-plus/Sandboxie/search?q=EmailReader
https://github.com/search?q=org%3Asandboxie-plus+4391&type=code
https://github.com/search?q=org%3Asandboxie-plus+4392&type=code

Note: about the latest two links, it's enough to update Sbie-English-1033.txt. I showed them to you in order to understand the connections.

Dyras:
I think I've added it everywhere now, aside from in the Sbie-English-1033.txt. Am I to assume I should use the number that comes after the last used one?

Sandboxie/apps/control/AppPage.cpp
```diff
@@ -1020,6 +1020,13 @@ void CAppPage::AddPages(CPropertySheet &sheet, const CString &BoxName)
     info.WithLink = FALSE;
     info.WithCreate = FALSE;
     m_app_pages.AddTail(new CAppPage(&info, BoxName));
+    
+    info.ClassName = L"MediaPlayer";
+    info.TitleId = MSG_4391;
+    info.LabelId = MSG_4392;
+    info.WithLink = FALSE;
+    info.WithCreate = FALSE;
+    m_app_pages.AddTail(new CAppPage(&info, BoxName));
```
isaak654, line 1029: 1) The two IDs are the same of EmailReader, so it would be better to create new IDs in Sbie-English-1033.txt. They contain a title and a detailed description displayed in Sbie Classic.
2) Considering that this class will be displayed in Sbie Classic, it will be rendered in a certain order. Please move the entire MediaPlayer block under Desktop class:
immagine

isaak654 (changes requested):

Dyras:
Aye aye sir, everything seems to be ready now!

Sandboxie/msgs/Sbie-English-1033.txt
```diff
@@ -3288,6 +3288,26 @@ Email Reader
...
+4393;txt;01
+Media Players
+.
+
+4394;txt;01
+The following exclusions allow Media players running in this sandbox to access files outside the sandbox.
+.
+
+4395;txt;01
+Force VLC Media Player to run in the Sandbox
```
isaak654, line 3300: Change it to:
Force VLC Media Player to run in this sandbox

isaak654 (changes requested):
> Aye aye sir, everything seems to be ready now!

In theory, yes. I never reviewed a new category before.

One last change:

Sandboxie/msgs/Sbie-English-1033.txt
```diff
@@ -3288,6 +3288,26 @@ Email Reader
...
+.
+
+4396;txt;01
+Allow VLC Media Player access to the VLC AppData folder
+.
+4397;txt;01
```
isaak654, line 3306: You need to start a new line before this string.

isaak654 (changes requested):

Sandboxie/msgs/Sbie-English-1033.txt
```diff
@@ -3288,6 +3288,26 @@ Email Reader
...
+Allow VLC Media Player access to the Photo folder for easier screenshotting
+.
+
```
isaak654, line 3309: Remove the additional space here.

Sandboxie/install/Templates.ini
```diff
@@ -2677,6 +2677,26 @@ OpenIpcPath=*\BaseNamedObjects*\{6560EAD3-F709-4B66-B90B-EA2D2C85AE3B}
...
+Tmpl.Title=#4396
+Tmpl.Class=MediaPlayer
+OpenFilePath=vlc.exe,%AppData%\vlc\*
+
+[Template_VLC_DirectAccess_Photos]
+Tmpl.Title=#4397
+Tmpl.Class=MediaPlayer
+OpenFilePath=vlc.exe,%USERPROFILE%\Pictures\*
+
+
```
isaak654, line 2699: I don't know if the placement would affect Sbie Classic order, but I would move it before the Download Managers block (just in case).

isaak654 (changes requested):

Dyras:
@DavidXanatos I got a template ready for SumatraPDF. Do you want that under Misc, Print or something else?

Print only seems to have actual printing, Misc seems to have nothing but compatibility. Or do you perhaps want a PDF class too?

isaak654:
> 
> 
> @DavidXanatos I got a template ready for SumatraPDF. Do you want that under Misc, Print or something else?
> 
> Print only seems to have actual printing, Misc seems to have nothing but compatibility. Or do you perhaps want a PDF class too?

In Classic, the Print section is displayed as PDF/Printing (as in the attachment below). Probably it would be more simple to rename the one in Plus (in my opinion).

image

Sandboxie/apps/control/AppPage.cpp
```diff
@@ -1041,6 +1041,13 @@ void CAppPage::AddPages(CPropertySheet &sheet, const CString &BoxName)
     info.WithLink = TRUE;
     info.WithCreate = FALSE;
     m_app_pages.AddTail(new CAppPage(&info, BoxName));
+    
+    info.ClassName = L"MediaPlayer";
+    info.TitleId = MSG_4393;
+    info.LabelId = MSG_4394;
```
isaak654, line 1047: Considering your latest changes, you can even replace MSG_4394 with the existent MSG_4207:
https://github.com/sandboxie-plus/Sandboxie/blob/d34ba4d9a6784629b1c5e3779d3a5848702555cb/Sandboxie/msgs/Sbie-English-1033.txt#L2909-L2910

and remove 4394;txt;01 from Sbie-English-1033.txt.

It's your call, it depends on which string you like more.

Sandboxie/msgs/Sbie-English-1033.txt
```diff
@@ -3288,6 +3288,18 @@ Email Reader
...
+4393;txt;01
+Media Players
+.
+
+4394;txt;01
+The following exclusions allow Media players running in this sandbox to access files outside the sandbox.
+.
+
+4395;txt;01
+Allow VLC direct access to the Photo folder for easier screenshotting.
```
isaak654, line 3300: I would replace 'for easier screenshotting' with 'for easier screen capture'.

isaak654 (changes requested):

Sandboxie/apps/control/AppPage.cpp
Dyras: Yeah I think you're right. The other message isn't awfully specific, IMO.

Dyras:
I've added everything from this into https://github.com/sandboxie-plus/Sandboxie/pull/719 so go look at that one instead :)

PR #741, opened 2021-04-01
Fixes qBittorrent and Transmission templates: AppDataLocal -> Local AppData

Dyras:
Yeah, so I screwed this up and this PR fixes the qBittorrent and Transmission templates. Sorry for the mistake.

Feel free adding this to the non-release version of 0.7.3.

### DIFF
--- a/Sandboxie/install/Templates.ini
+++ b/Sandboxie/install/Templates.ini
@@ -2761,7 +2761,7 @@ ForceProcess=qBittorrent.exe
 [Template_qBittorrent_DirectAccess_Profile]
 Tmpl.Title=#4338,qBittorrent
 Tmpl.Class=TorrentClient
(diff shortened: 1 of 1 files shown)


isaak654 (changes requested):
`%LocalAppData%` is considered incorrect for Sandboxie, so the correct variable is `%Local AppData%`, it's specified in the list of variables in the [documentation](https://davidxanatos.github.io/sandboxie-docs/Content/ExpandableVariables.html).

You can also see an example here:
https://github.com/sandboxie-plus/Sandboxie/blob/bdea28a8a3be498540bb28c450ec515c0912c2f0/Sandboxie/install/Templates.ini#L133

Dyras:
Well I be damned. %LocalAppdata% actually works for me.

Issue #536, opened 2021-02-07
Videolan vlc.exe tries to write uneeded copy of "files.mp4" inside the box


Rngexile:
I've just noticed the same thing as someone in 
https://www.wilderssecurity.com/threads/released-sandboxie-plus-sbie-fork-versions-with-signed-driver.434924/page-34

1/ When launched outside sandboxie, the media player vlc.exe (v3.12 https://www.videolan.org/) can read video files (examples) AAA.mp4 and a renamed copy AAA.mp4.mkv (or avi or maybe other containers) without the need to (re)write these files.

I can see in procmonitor (from Sysinternals) that vlc.exe only needs "**Desired Access: Generic Read**" on the file AAA.mp4.mkv 
and needs at some points "**Desired Access: Generic Read/Write,** Disposition: Open, Options: Synchronous IO Non-Alert, Non-Directory File, Attributes: n/a, **ShareMode: Read**, AllocationSize: n/a, OpenResult: Opened" for the file AAA.mp4.

Still, no (re)writes of AAA.mp4, no change on the file. Expected behaviour.

2/ **But, when vlc is launched inside his box, it seems sandboxie would like to copy (and succeed if copy size limit is high enough) the file AAA.mp4 inside the box.**
If AAA.mp4 has higher size than the configurer copy size limit, then sandboxie ask us if we allow the copy of the file (even if the ask message 2102 is disabled)
If we don't allow the copy, vlc.exe reads AAA.mp4 like normal (like a read only file).

AAA.mp4.mkv is read by vlc.exe in his box without the will of making a copy inside the box.

Is it a normal behaviour from sandboxie because of the "Desired Access: Generic Read/Write" asked by vlc for mp4 files (maybe it's a vlc's fault ?) ?
The copy, even for small files is uneeded.

Or maybe is it relative to the bug "File Copy Limit wraps at 4GB #526" ?

OS : windows10 x64 (all updates)
VLC : 3.12 x64
Sandboxie-Plus : 0.67

Thank you.

DavidXanatos:
I would recommend to add MP4's to the list of files that are never to be migrated
see recent changelog:

- reworked file migration behaviour, removed hardcoded lists in favour of templates
-- you can now use "CopyAlways=", "DontCopy=" and "CopyEmpty=" that support the same syntax as "OpenFilePath="
-- "CopyBlockDenyWrite=program.exe,y" makes a write open call to a file that won't be copied fail instead of turning it read-only

so add
[...]

PR #743, opened 2021-04-01
Prevent media players from importing video files

Dyras:
Hello! This prevents VLC and PotPlayer from importing media files if the template is used.

Until a better fix is available, this fixes https://github.com/sandboxie-plus/Sandboxie/issues/536

### DIFF
--- a/Sandboxie/install/Templates.ini
+++ b/Sandboxie/install/Templates.ini
@@ -2632,6 +2632,11 @@ OpenWinClass=ZT9MainWindow
 Tmpl.Title=#4323,VLC
 Tmpl.Class=MediaPlayer
 ForceProcess=vlc.exe
+DontCopy=vlc.exe,*.mp4
+DontCopy=vlc.exe,*.mkv
+DontCopy=vlc.exe,*.avi
+DontCopy=vlc.exe,*.flv
+DontCopy=vlc.exe,*.wmv
 
 [Template_VLC_DirectAccess_Profile]
 Tmpl.Title=#4338,VLC
--- a/Sandboxie/install/Templates.ini
+++ b/Sandboxie/install/Templates.ini
@@ -2628,15 +2628,20 @@ OpenWinClass=ZT9MainWindow
 # Media Players
 #
 
+[Template_No_Importing_Of_Media_Files]
+Tmpl.Title=#4399
+Tmpl.Class=MediaPlayer
+DontCopy=*.mp4
+DontCopy=*.mkv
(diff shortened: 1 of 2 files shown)


Dyras:
Any thoughts on this approach is appreciated @DavidXanatos

DavidXanatos:
Eider add these for all the known players, or a global tamplate that prevents copying of media files would be better

Dyras:
@DavidXanatos I've added a template for this now. Going to add mp3 and the like later.

Sandboxie/install/Templates.ini
```diff
@@ -2628,11 +2628,26 @@ OpenWinClass=ZT9MainWindow
...
+DontCopy=*.mkv
+DontCopy=*.avi
+DontCopy=*.flv
+DontCopy=*.wmv
+DontCopy=*.mp3
+DontCopy=*.aac
+DontCopy=*.m4a
+DontCopy=*.flac
+DontCopy=*.ogg
+
```
isaak654, line 2644: Consider adding .wav, .wma, .ape audio extensions.

isaak654 (changes requested):

Sandboxie/install/Templates.ini
```diff
@@ -2628,6 +2628,23 @@ OpenWinClass=ZT9MainWindow
...
+DontCopy=*.wmv
+DontCopy=*.mp3
+DontCopy=*.aac
+DontCopy=*.m4a
+DontCopy=*.flac
+DontCopy=*.ogg
+DontCopy=*.wav
+DontCopy=*.wma
+DontCopy=*.ape
+
```
isaak654, line 2647: Consider to add .mpg, .mpeg, .mov, .webm, .f4v, .rm, .swf, .ts or other video formats listed here: 
https://filesamples.com/categories/video
You can do the same with audio formats like .ac3, .oga, .aiff, .ra, .cdda, or others: 
https://filesamples.com/categories/audio

isaak654 (changes requested):

Dyras:
Unless there is something else that I've missed, I think that's all.

isaak654:
Note: `DontCopy=*.ts` was recently commented out from Templates.ini in commit https://github.com/sandboxie-plus/Sandboxie/commit/a21d41af17a46a66c721b602515fee6d00006b58 because it breaks Cyberpunk 2077 compatibility. However, you can still add it in your Sandboxie.ini configuration to make it permanent.

PR #778, opened 2021-04-12
Display PDF/Print category in Plus UI

isaak654:
@DavidXanatos I need a review about the validity of this commit. 
The scope of this commit is to rename the Print category to PDF/Print (only in Plus UI).

### DIFF
--- a/SandboxiePlus/QSbieAPI/Sandboxie/SbieTemplates.cpp
+++ b/SandboxiePlus/QSbieAPI/Sandboxie/SbieTemplates.cpp
@@ -258,7 +258,7 @@ void CSbieTemplates::CollectTemplates()
 
 	QStringList Templates;
 	Templates.append(GetTemplateNames("EmailReader"));
-	Templates.append(GetTemplateNames("Print"));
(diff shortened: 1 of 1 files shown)


isaak654:
@DavidXanatos I need a review about this commit. 
The scope of this commit is to rename the Print category to PDF/Print (only in Plus).

isaak654:
@DavidXanatos I need a review about the validity of this commit. 
The scope of this commit is to rename the Print category to PDF/Print (only in Plus UI).

DavidXanatos:
GetTemplateNames returns a string list you can not use "bla" + string list
The goal is to change the category name in the UI as displayed in the template list and the filter combo box right?

isaak654:
> 
> 
> GetTemplateNames returns a string list you can not use "bla" + string list
> The goal is to change the category name in the UI as displayed in the template list and the filter combo box right?

Yeah, this is needed **only** in the Plus UI, not in Classic UI.
It would allow to avoid to replace the Print variable in the `Tmpl.Class=Print` entries listed in templates.ini.

PR #788, opened 2021-04-15
Add Tmpl.Scan options for anti-phishing rules

isaak654:
With this commit, automatic detection of anti-phishing rules per browser is achieved through appropriate scan options, so it deprecates the use of `Firefox_Phishing_DirectAccess` and `Chrome_Phishing_DirectAccess` as default templates.

By approving this commit, the decision to accept these templates is demanded to the user after the upgrade through the Software Compatibility tab (the `#4337,` reference will be replaced with 'Allow direct access to xxx phishing database'):
Software_compatibility_tab

I tested successfully the detection of all browsers supported by Sandboxie, including Firefox and Edge.
The first one was a bit hard to be recognized automatically in Classic's Software Compatibility, but it was sufficient to apply and uncheck twice `Firefox_Phishing_DirectAccess` in DefaultBox settings. 

New configurations will benefit from this change, but also all old configurations cleaned from the `Firefox_Phishing_DirectAccess` and `Chrome_Phishing_DirectAccess` references inside sandboxes.


APMichael:
JFYI: In the past, the asterisk "\*" after "safebrowsing" was necessary. The update was done via temporary files or folders. Without the asterisk "\*", the update then always failed.

isaak654:
@APMichael
Ok, thank your for the info. I don't want to take any risk, so I'll edit the commit accordingly.

APMichael:
Yes, "blocklist.xml" and "cert9.db" work fine. These do not need an asterisk.

isaak654:
I tested extensively the detection for all browsers, so this pull request is not a working in progress anymore.
Now it's up to @DavidXanatos, but obviously any review is appreciated.

isaak654:
Nobody seems interested in this PR, so it's time to close.

If anyone wants, just let me know here and I'll recreate these changes in a new PR.

PR #905, opened 2021-06-14
Refine message strings, updating all .ts files and including Simplified Chinese fixes

yfdyh000:
By the way, correct some letters case and missing strings for i18n.

In addition, I have tried to translate some names in CSbieProcess, but worry that some experts are more familiar with their English names.


DavidXanatos:
The changes seam not to build properly can you please fix it.

isaak654:
The changes you made in OptionsWindow.ui are already in #904, so I'm going to remove the mentioned file from this pull request in order to avoid troubles with it.

isaak654:
@yfdyh000 Is there a script to fix the lines with outdated line references and add new entries available only in english to the TS file?

yfdyh000:
> @yfdyh000 Is there a script to fix all .ts files with outdated line references and adding localization support for the new entries available only in english?

@isaak654 Run the `C:\Qt\6.1.1\msvc2019_64\bin\lupdate.exe SandMan.pri` at `\SandboxiePlus\SandMan`.

PR #914, opened 2021-06-16
Fix buttons position in Classic UI settings

isaak654:
This is usually a problem with some translations, so I made a few changes.


isaak654:
@DavidXanatos 
I updated this commit with the changes introduced in the successful sample file that I sent you yesterday.
For me it's ready.

PR #929, opened 2021-06-19
Adds a LibreWolf template

Dyras:
This adds a template for LibreWolf. Haven't tested it yet, seeing as LibreWolf isn't working in Sandboxie-Plus at the time of writing.


isaak654 (changes requested):
As you can see [here](https://github.com/sandboxie-plus/Sandboxie/blob/712617c13cf91c45f3271779977ea423b3250485/Sandboxie/core/dll/dllmain.c#L574), proper LibreWolf support was recently added in the source code. 
Now you only need to add `|| (name.Find(L"LibreWolf_") != -1);` under this line (to display the template in Classic UI):
https://github.com/sandboxie-plus/Sandboxie/blob/master/Sandboxie/apps/control/AppPage.cpp#L548

PS: please make sure to erase `;` symbol from SeaMonkey when you do the change.

isaak654 (approved):

PR #965, opened 2021-07-06
Update Changelog 5.50.5-5.50.6

NewKidOnTheBlock:
Thank you for your contribution to the Sandboxie repository.

In order to reduce the risk of accidental merges, it's highly recommended for beginners to open a new Pull Request in draft state by clicking the button "Create Draft Pull Request", instead of using the default "Create Pull Request". You can convert an existing pull request to a draft by clicking the "Convert to draft" link in the right sidebar under "Reviewers".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,16 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 
+## [0.8.6 / 5.50.6] - 2021-07-07
+
+### Added
+- added LibreWolf template (by Dyras) [#929](https://github.com/sandboxie-plus/Sandboxie/pull/929)
+
+### Fixed
+- fixed performance bug introduced in 0.8.5
+
+
+
 ## [0.8.5 / 5.50.5] - 2021-07-06
 
 ### Added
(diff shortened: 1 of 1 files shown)


NewKidOnTheBlock:
If you had told me to press the Break key on my keyboard, I would not have found it.
This might be particular to German keyboards, but I'm pretty sure it's also called Pause key on American/English ones.

isaak654:
> 
> 
> If you had told me to press the Break key on my keyboard, I would not have found it.
> This might be particular to German keyboards, but I'm pretty sure it's also called Pause key on American/English ones.

Break is an english word, so it can be found in US layouts:
https://www.computerhope.com/jargon/b/break.htm

In some keyboards with italian layouts, I have Pausa / Interr, while in others just Pausa.

Interr is the translation of break in my language.

NewKidOnTheBlock:
An image search revealed that US-keyboards either call it "Pause" or "Pause | Break" (= both inscriptions on the key)
On my German keyboard it's called "Pause | untbr".

isaak654:
Pause | Break seems fine to me, it covers all the cases.

isaak654 (changes requested):
Dialog is preferred instead of Dialogue, because it has a specific use in computational contexts.

isaak654 (approved):

PR #1021, opened 2021-07-20
Update sandman_es.ts

sebadamus:
First attemt to contribute from github web... lets see... @isaak654 is it ok?
(created as draft pull as recommended by github web)

### DIFF
--- a/SandboxiePlus/SandMan/sandman_es.ts
+++ b/SandboxiePlus/SandMan/sandman_es.ts
@@ -509,21 +509,21 @@
     </message>
     <message>
         <location filename="Windows/PopUpWindow.cpp" line="229"/>
-        <source>Do you want to allow %4 (%5) to copy a %1 large file into sandbox: %2?<byte value="xd"/>
+        <source>Do you want to allow %4 (%5) to copy a %1 large file into sandbox: %2?
 File name: %3</source>
         <translation>Queres permitir %4 (%5) copiar a %1 archivo grande dentro de la sandbox: %2
 Nombre de archivo: %3</translation>
     </message>
     <message>
         <location filename="Windows/PopUpWindow.cpp" line="235"/>
-        <source>Do you want to allow %1 (%2) access to the internet?<byte value="xd"/>
+        <source>Do you want to allow %1 (%2) access to the internet?
 Full path: %3</source>
         <translation>Queres permitir %1 (%2) acceder a internet?
 Ruta completa: %3</translation>
     </message>
     <message>
         <location filename="Windows/PopUpWindow.cpp" line="298"/>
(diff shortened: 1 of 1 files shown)


isaak654:
I would suggest to restore the `<byte value="xd"/>` references, in my experience they are needed in certain points to show the translations correctly in the program. Without them, certain strings will be only displayed in english.

As you can read here, other .ts languages use them as well:
https://github.com/sandboxie-plus/Sandboxie/search?q=%3Cbyte+value%3D%22xd%22%2F%3E

You just need to create another commit to your existent patch-2 branch on your forked repo and it will be automatically attached here, on this pull request.

sebadamus:
> 
> 
> I would suggest to restore the `<byte value="xd"/>` references as they were, in my experience they are needed in certain points to show the translations correctly in the program. Without them, certain strings will be only displayed in english.
> 
> As you can read here, other .ts languages use them as well:
> https://github.com/sandboxie-plus/Sandboxie/search?q=%3Cbyte+value%3D%22xd%22%2F%3E
> 
> ~You just need to create another update commit to your existent patch-2 branch on your forked repo and it will be automatically attached here, on this pull request.~ This is often useful, instead to open another pull request (having one place to look at changes on the same file is always preferred).

Hi @isaak654 sorry for the delay, I am now used to github desktop/web.  

1) Can you please tell me if the "commit" I have done last month was ok? so I continue...

2) I almost understood what you told me about the <byte value="xd"/>, can you clarify? Did you mean that on every statement like this:

**<source>**text text text here more text### <byte value="xd"/>**</source>**

The <byte value="xd"/> must be just before this </source>?

I did not delete those <byte value="xd"/>, maybe it got deleted after my copy/paste, but can restore and commit again if you clarify me that!

Thanks and sorry for my ignorance.

isaak654:
@sebadamus 
Don't worry, I already fixed this pull request for you and merged it on the source code 22 days ago (see the history of changes above):
immagine

Other things need to be done, like adding the translation of Plus installer in Spanish (only in the CustomMessages section):
https://github.com/sandboxie-plus/Sandboxie/blob/master/Installer/Languages.iss

If you need more details about how to update your `sandman_es.ts` constantly at 100% without leaving any entry untranslated, you could run [this](https://github.com/sandboxie-plus/Sandboxie/pull/905#issuecomment-882049888) command line manually after a major release with new UI changes. Unfortunately that command line requires the installation of [QT Online Installer](https://www.qt.io/download) (the opensource version is at the bottom of the page) or there is the [5.12.x Offline Installer](https://www.qt.io/offline-installers) that is stuck at version 5.12.11.

I don't know if there is a UI option to do that without running the command in a command prompt, surely I didn't find any option in the standalone Linguist program.

PR #1031, opened 2021-07-23
Update CHANGELOG.md

isaak654:


isaak654:
@NewKidOnTheBlock 
Could you help me to review it, please?

NewKidOnTheBlock:
Man, it was a hassle just to get the latest file that includes the latest edits.
Since I've got no idea how I'm supposed to do this in github, here is my edit:

## [0.9.0 / 5.51.0] - 2021-07-??

### Added
- added support for Windows Filtering Platform (WFP) to be used instead of the device-based network blocking scheme
-- to enable this support, add 'NetworkEnableWFP=y' to the global section and reboot or reload the driver
-- to use WFP for a specified sandbox, add 'AllowNetworkAccess=n' to said box
-- you can exempt certain processes from blocking by using 'AllowNetworkAccess=program.exe,y'
-- you can also enable this policy globally by adding 'AllowNetworkAccess=n' to the global section
-- in this case you can exempt entire boxes by adding 'AllowNetworkAccess=n' to said boxes
-- specifying 'AllowNetworkAccess=program.exe,n' will block the access only for the named process
-- Note: WFP is less absolute than the old approach, using WFP will filter only TCP/UDP communication
--	restricted boxed processes will still be able to resolve domain names using the system service
--  however, they will not be able to send or receive data packets directly
-- the advantages of WFP is that filter rules can be implemented by restricting communication only to specified addresses or selected ports using "NetworkAccess=..."
- added fully functional rule-based packet filter in user mode for the case when "NetworkEnableWFP=y" is not set
-- the mechanism replaces the old "BlockPort=..." functionality
-- Note: this filter applies only to outgoing connections/traffic, for incoming traffic either the WFP mode or a third-party firewall is needed
-- like the old user mode based mechanism, malicious applications can bypass it by unhooking certain functions
-- hence it's recommended to use the kernel mode WFP-based mechanism when reliable isolation is required
-- Note: the main reason this mechanism was added in user mode is to allow for easier debugging
- added new trace option "NetFwTrace=*" to trace the actions of the firewall components
-- please note that the driver only trace logs the kernel debug output, use DbgView.exe to log
- API_QUERY_PROCESS_INFO can now be used to get the impersonation token of a sandboxed thread
-- Note: this capability is used by TaskExplorer to allow inspecting sandbox-internal tokens
-- Note: a process must have administrative privileges to be able to use this API
- added a UI option to switch "MsiInstallerExemptions=y" on and off
-- just in case, if a future Windows build breaks something in the systemless mode
- added sample code for ObRegisterCallbacks to the driver
- added new debug options "DisableFileFilter=y" and "DisableKeyFilter=y" that allow to disable file and registry filtering
-- Note: these options are for testing only and disable core parts of the sandbox isolation
- added a few command line options to SandMan.exe

### Changed
- greatly improved the performance of the trace log, but it's no longer possible to log both SandMan and SbieCtrl at the same time
- changed code integrity verification policies
-- code signature validation of user mode components is disabled when Windows is booted in test-signing mode
- reworked process creation code to use PsSetCreateProcessNotifyRoutineEx and improved process termination

### Fixed
- added missing hook for ConnectEx function



## [0.8.9 / 5.50.9] - 2021-07-??

### Changed
- updated a few icons
- updated GitHub build action to use Qt 5.15.2
- improved the "full" tray icon to be more distinguishable from the "empty" one
- changed code integrity verification policies
-- code signature is no longer required to change config, to protect presets use the existing "EditAdminOnly=y"

### Fixed
- fixed issue with systemless MSI mode introduced in the last build
- fixed MSI installer not being able to create action server on Windows 11
- fixed MSI installer not working in systemless mode on Windows 11

isaak654:
> The 2nd to last sentence sounds wierd:
> `- fixed MSI installer not being able to create action server on Windows 11`
> Either "action server**s**" (=plural) 
> or
> put in an article: "create **an** action server"

Days ago he posted "so 11 broke the action server mechanism", so I updated the changelog accordingly.
Thank you for the help!

PR #1035, opened 2021-07-24
Update Dutch translations

RockyTDR:
Update Dutch translations for Sandboxie Classic and Sandboxie-Plus

### DIFF
--- a/SandboxiePlus/SandMan/sandman_nl.ts
+++ b/SandboxiePlus/SandMan/sandman_nl.ts
@@ -648,21 +648,21 @@
         <location filename="Windows/PopUpWindow.cpp" line="253"/>
         <source>Do you want to allow %4 (%5) to copy a %1 large file into sandbox: %2?<byte value="xd"/>
 File name: %3</source>
-        <translation>Wilt u %4 (%5) toestaan om een %1 groot bestand te kopiëren naar sandbox: %2?
+        <translation>Wilt u %4 (%5) toestaan om een %1 groot bestand te kopiëren naar sandbox %2?&lt;byte value=&quot;xd&quot;/&gt;
 Bestandsnaam: %3</translation>
     </message>
     <message>
         <location filename="Windows/PopUpWindow.cpp" line="259"/>
         <source>Do you want to allow %1 (%2) access to the internet?<byte value="xd"/>
 Full path: %3</source>
-        <translation>Wilt u toestaan dat %1 (%2) toegang tot het internet heeft?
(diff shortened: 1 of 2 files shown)


SandboxiePlus/SandMan/sandman_nl.ts
```diff
@@ -648,21 +648,21 @@
         <location filename="Windows/PopUpWindow.cpp" line="253"/>
         <source>Do you want to allow %4 (%5) to copy a %1 large file into sandbox: %2?<byte value="xd"/>
 File name: %3</source>
-        <translation>Wilt u %4 (%5) toestaan om een %1 groot bestand te kopiëren naar sandbox: %2?
+        <translation>Wilt u %4 (%5) toestaan om een %1 groot bestand te kopiëren naar sandbox %2?&lt;byte value=&quot;xd&quot;/&gt;
```
isaak654, line 651: Byte value here has a strange format, it looks different than the `<byte value="xd"/>` that you can find two lines above.

Same goes for the following lines below: 658, 665, 688, 1054-1055 and 1135-1136

isaak654 (changes requested):

SandboxiePlus/SandMan/sandman_nl.ts
isaak654: I think I understood the issue... you don't need any byte value inside the translation tags. The code was only required for source tags.

Please remove those references that you're trying to add.

RockyTDR: Done! `Sandman_nl.ts` now only has a colon changed. The other translation edits were in the Sandboxie Classic. 

Note that other translators will also see (and translate) the byte codes because they are part of the source string.

isaak654:
> 
> 
> Done! `Sandman_nl.ts` now only has a colon changed. The other translation edits were in the Sandboxie Classic.
> 
> Note that other translators will also see (and translate) the byte codes because they are part of the source string.

@RockyTDR

Unfortunately it's the only workaround we know at this moment to force the translation of single english strings with multiple lines in other languages. For example, this is what appears without the byte codes: recovery_window

Any other alternative would be welcome.

RockyTDR:
> 
> 
> > Done! `Sandman_nl.ts` now only has a colon changed. The other translation edits were in the Sandboxie Classic.
> > Note that other translators will also see (and translate) the byte codes because they are part of the source string.
> 
> @RockyTDR
> 
> Unfortunately it's the only workaround we know at this moment to force the translation of single english strings with new lines in other languages. For example, this is what appears without the byte codes: recovery_window
> 
> Any other alternative would be welcome.

Ok. Thanks for the info! Can this PR be approved?

isaak654 (approved):

PR #1037, opened 2021-07-25
Fix innosetup script to remove shell integration keys

mpheath:
Remove shell integration in innosetup fix with `RegDeleteKeyIncludingSubkeys` function, instead of `RegDeleteValue` function as `sandbox` is a registry **key**, not a registry **value**. `RegDeleteValue` would have been failing (*with a return value of False*) as it cannot remove keys and the subkeys as it is [documented](https://jrsoftware.org/ishelp/index.php?topic=isxfunc_regdeletevalue) to remove a value.

These are the `sandbox` keys copied from RegEdit that require keys and subkeys removed at uninstall of Sandboxie Plus.

```
HKEY_CURRENT_USER\Software\Classes\*\shell\sandbox
HKEY_CURRENT_USER\Software\Classes\Folder\shell\sandbox
```

This fix has not been compiled and tested, though it should be an improvement. No other registry functions appear to have any similar issues with a quick review of the innosetup source.

### DIFF
--- a/Installer/Sandboxie-Plus.iss
+++ b/Installer/Sandboxie-Plus.iss
@@ -404,8 +404,8 @@ begin
   RegDeleteValue(HKEY_CURRENT_USER, 'Software\Microsoft\Windows\CurrentVersion\Run', 'SandboxiePlus_AutoRun');
   
   // remove shell integration
-  RegDeleteValue(HKEY_CURRENT_USER, 'software\classes\*\shell', 'sandbox');
-  RegDeleteValue(HKEY_CURRENT_USER, 'software\classes\folder\shell', 'sandbox');
+  RegDeleteKeyIncludingSubkeys(HKEY_CURRENT_USER, 'software\classes\*\shell\sandbox');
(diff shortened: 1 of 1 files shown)


isaak654:
What do you think about adding these?
```
RegDeleteKeyIncludingSubkeys(HKEY_CLASSES_ROOT, '*\shell', 'sandbox');
RegDeleteKeyIncludingSubkeys(HKEY_CLASSES_ROOT, 'folder\shell', 'sandbox');
```

I'm not sure if they were part of an old Plus install...

mpheath:
HKEY_CLASSES_ROOT is a mirror of combined HKCU and HKLM. It should not be written to. Same could be said for removal. Writing to HKCR can overwrite whatever exists in HKCU or HKLM as you do not know at write time. It would not be recommended to handle, in my opinion. You may expect it to do HKCU, but if not there and HKLM entries exist, they may be removed even though they should remain. It may not do harm in this case.

isaak654:
> HKEY_CLASSES_ROOT is a mirror of combined HKCU and HKLM.

For some reason I forgot that detail, probably because other scripts try to fiddle with it.
Imho this pull request should be merged in the next few hours.

PR #1078, opened 2021-08-07
Update Chinese Traditional

Hulen:
Update Chinese Traditional


isaak654:
You need to replace all `ChineseSimplified.` references with `ChineseTraditional.`

PR #1079, opened 2021-08-07
Update Languages.iss

bastik-1001:
Added German translation for the multilanguage installer.


isaak654:
Do you have the chance to merge all those commits into one?

bastik-1001:
I'm sorry, I don't see how I could squash the commits as I did not notice an option to do so before and now I don't see an option.

I updated the fork, before applying the changes to a single file, but it is more than one commit, even though that would not need to be the case. I just did not see a way to select only one of the commits to be merged or making them into one.

PR #1107, opened 2021-08-14
Update sandman_zh-TW.ts

Hulen:
Update sandman_zh-TW


isaak654:
`<byte value="xd"/>` is needed **only** for single source tags that contain multiple new lines, and it's not intended to be used inside the translation tags. Currently this is the only workaround that communicates to Sandboxie-Plus UI (based on Qt platform) to recognize those specific lines correctly in all other languages than english.

isaak654:
There are some conflicts with the changes introduced by this translator: https://github.com/sandboxie-plus/Sandboxie/pull/1104

So this pull request can't be merged at all, unless you fix the conflicts.

isaak654 (changes requested):
There are some conflicts with the changes introduced by this translator: https://github.com/sandboxie-plus/Sandboxie/pull/1104

So this pull request can't be merged at all, unless you fix the conflicts.

PR #1144, opened 2021-08-28
Update SRWare Iron template to actually work

Dyras:
So for whatever reason, SRWare Iron has gone from calling their program "iron.exe" to "chrome.exe"

There's still an iron.exe in the folder but it doesn't seem to be used.

This changes the force file template to force the install folder in Sandboxie instead. The alternative is to force "chrome.exe" in the sandbox, which will be confusing for people that use both Chrome and Iron.

"chrome.exe" is also used instead of "iron.exe" for every occurrence. 

Iron doesn't seem to have an updater, so I expect no issues with that. I installed it with Chocolatey.

### DIFF
--- a/Sandboxie/install/Templates.ini
+++ b/Sandboxie/install/Templates.ini
@@ -533,7 +533,8 @@ OpenFilePath=dragon.exe,%Tmpl.Dragon%\Bookmarks*
 OpenFilePath=dragon.exe,%Tmpl.Dragon%\Favicons*
 
 [Template_Dragon_History_DirectAccess]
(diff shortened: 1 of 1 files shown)


isaak654 (changes requested):
Fix spaces on 536-537 and 583-584 lines

isaak654:
I'm having second thoughts about adjusting this pull request and create a new commit just to see fixed spaces (one of them related to another template) in the commit's history. I think it would be better to create a new pull request from start.

Dyras:
What the hell, no idea where those lines came from

PR #1146, opened 2021-08-28
Yet another attempt at fixing SRWare Iron

Dyras:
<div type='discussions-op-text'>

So for whatever reason, SRWare Iron has gone from calling their program "iron.exe" to "chrome.exe"
\
There's still an iron.exe in the folder but it doesn't seem to be used.
\
This changes the force file template to force the install folder in Sandboxie instead. The alternative is to force "chrome.exe" in the sandbox, which will be confusing for people that use both Chrome and Iron.
\
"chrome.exe" is also used instead of "iron.exe" for every occurrence. 
\
Iron doesn't seem to have an updater, so I expect no issues with that. I installed it with Chocolatey.</div>

There shouldn't be any extra lines now. If there is, I didn't make them. GitHub is weird man.

### DIFF
--- a/Sandboxie/install/Templates.ini
+++ b/Sandboxie/install/Templates.ini
@@ -579,54 +579,54 @@ Tmpl.Class=WebBrowser
 [Template_Iron_Force]
 Tmpl.Title=#4323,SRWare Iron
 Tmpl.Class=WebBrowser
-ForceProcess=iron.exe
+ForceFolder=C:\Program Files\SRWare Iron (64-Bit)
 
 [Template_Iron_Bookmarks_DirectAccess]
 Tmpl.Title=#4356,SRWare Iron
 Tmpl.Class=WebBrowser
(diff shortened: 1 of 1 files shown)


isaak654:
GitHub can act weird when it attempts to converts line endings. 
Things are improved for me after using GitHub Desktop with `autocrlf = false` on my `.gitconfig` file (located in `C:\Users\MyUsername`):

.gitconfig

Do you use the GitHub web editor or a specific offline editor?

Dyras:
I use the website. Might check out the client later.

isaak654:
Your findings related to SRWare Iron were also reported on the SRWare Iron Forum: [iron.exe vs chrome.exe](https://www.srware.net/forum/viewtopic.php?t=7732)

I also tested the changes introduced in this pull request and I think they're fine.

isaak654 (approved):

PR #1148, opened 2021-08-30
Update Changelog 5.51.5

NewKidOnTheBlock:
We need to talk about your T.P.S. reports.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [0.9.5 / 5.51.5] - 2021-08-30
 
 ### Added
-- added option to run a sandbox in session 0
+- added option to run a sandbox in [session 0](https://techcommunity.microsoft.com/t5/ask-the-performance-team/application-compatibility-session-0-isolation/ba-p/372361)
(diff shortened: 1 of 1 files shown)


isaak654:
I moved the system tray option with the Classic icon from the Changed section to the Added section.

NewKidOnTheBlock:
What does "session 0" mean?
The orthography of the changelog should be alright now, but if you asked me to explain what "session 0" is, I wouldn't be able to give an answer.

isaak654:
I think these explanations could help:
https://techcommunity.microsoft.com/t5/ask-the-performance-team/application-compatibility-session-0-isolation/ba-p/372361
https://github.com/sandboxie-plus/Sandboxie/blob/4efa6fd4f81fd6795700cdb7ca734bf6111388bb/Sandboxie/core/svc/ProcessServer.cpp#L1093-L1101

NewKidOnTheBlock:
I've added the explanation as a hyperlink.

DavidXanatos:
In windows you have so called logon sessions, when a user loggs in a new session is created for him and all his programs are associated with this session.
Now what do programs do that start before any user loggs in like services or winlogon or etc....
well windows has a session 0 for that its where all the services and otehr system processes live which do not interact with the user.

isaak654:
@NewKidOnTheBlock
I made another commit with a small change, let me know if it works for you.

PR #1215, opened 2021-09-19
[Plus only] Disable Turkish for inactivity

isaak654:
There is a concrete risk of a translation mismatch because of the continuos Plus UI changes. Since the latest Turkish update, more than 20 versions were released, so the Plus support for Turkish should be turned off.

If you want to update and restore the translation for Sandboxie Plus, please open a new issue.


mpheath:
The disabling in *Languages.iss* is OK. I am not sure about the parsing of *SandMan.pri*. Those backslashes at the end of the lines are continuation characters. The backslash and newline sequence will be stripped and the next line will be joined with the previous line.

So this effect may happen when parsed:
 
`    sandman_ru.ts     sandman_tr.ts #   sandman_tr.ts \ (unmaintained)    sandman_zh-CN.ts     sandman_zh-TW.ts`

I would suggest to move the commented `sandman_tr.ts` down below `sandman_zh-TW.ts` so to avoid issues with the line continuations.

If the parser strips the comment lines, then strips the line continuations and joins the lines, then it might be OK. That I am unsure if it will do that without testing it.

mpheath (approved):

PR #1217, opened 2021-09-21
Updated Russian translation

lufog:


isaak654:
Could you translate these lines too in [Languages.iss](https://github.com/sandboxie-plus/Sandboxie/blob/master/Installer/Languages.iss)?
https://github.com/sandboxie-plus/Sandboxie/blob/c78c60bf7a0bba4ee5660f1b4d3d3f68fbfc0b35/Installer/Languages.iss#L188-L192

PR #1256, opened 2021-10-06
Update bug_report.md

isaak654:


hg421:
> @hg421 It should be fine now, my next move is to update this template with the steps required to take a log (for both Plus and Classic). For example, something like [this](https://git.io/JwfHE), may I ask your opinion about it?

Yes it's definitely a good idea to have step for step instructions for that, on the other hand there are also bug reports which don't need a trace log, and I woudn't make this template too long.
Alternatively we could put the steps somewhere in the docs and just add a link here maybe

Issue #1155, opened 2021-08-31
No French translation


clexanis:
**Is your feature request related to a problem? Please describe.**
Sandboxie-Plus has no French translation.

**Describe the solution you'd like**
I'm currently translating Sandboxie-Plus into French.

**Describe alternatives you've considered**
As I have started a translation, I am opening this ticket to inform potential other users that a translation is in progress, and to avoid having duplicate tasks.

**Additional context**
 The translation in progress can be found here: [clexanis/Sandboxie:translation_fr](https://github.com/clexanis/Sandboxie/tree/translation_fr)
I will open a merger application when it is close to completion.

---

### Current progress

- [ ] Add reference to the translation

**Items to translate:**
- [x] CMultiErrorDialog
- [ ] CNewBoxWindow
- [ ] COptionsWindow
- [x] CPopUpMessage
- [ ] CPopUpProgress
- [ ] CRecoveryWindow
- [ ] CSandBoxPlus
- [ ] CSandMan
- [ ] CSbieModel
- [ ] CSbieProcess
- [ ] CSbieView
- [ ] CSelectBoxWindow
- [ ] CSnapshotsWindow
- [ ] CTraceModel
- [ ] CTraceView
- [ ] NewBoxWindow
- [ ] OptionsWindow
- [ ] PopUpWindow
- [ ] QObject
- [ ] QPlatformTheme
- [ ] RecoveryWindow
- [x] SelectBoxWindow
- [ ] SettingsWindow
- [ ] SnapshotsWindow

isaak654:
In case you want to translate Plus installer, see [here](https://github.com/sandboxie-plus/Sandboxie/blob/master/Installer/Languages.iss).

If you need additional tips, see https://github.com/sandboxie-plus/Sandboxie/discussions/1123#discussioncomment-1203489.

clexanis:
> 
> 
> In case you want to translate Plus installer, see [here](https://github.com/sandboxie-plus/Sandboxie/blob/master/Installer/Languages.iss).
> 
> If you need additional tips, see [#1123 (comment)](https://github.com/sandboxie-plus/Sandboxie/discussions/1123#discussioncomment-1203489).

I added the installer to the to-do list.
And, thanks for the tips.

isaak654:
I just wanted to inform you that new strings related to the Plus installer are available in [this pull request](https://github.com/sandboxie-plus/Sandboxie/pull/1058/files) and they are ready to be translated.

[...]

PR #1282, opened 2021-10-15
Add French translation

clexanis:
This pull is a follow-up to trouble ticket #1155.

The translation is mostly complete, but it has not yet been tested in detail.

#### Things to-do:
- [x] Sandboxie/Installer/Sandboxie-Plus.iss

> Note: As this pull includes several unnecessary commits, I think it would be more interesting to squash it.


isaak654:
Thank you for adding the missing line, I think this pull request is ready to be merged (with a Squash and merge).
If there is something else you would like to update, you can release a new pull request.

PR #1301, opened 2021-10-18
Made the list header text "Sandbox" in SelectBoxWindow translatable and added Chinese Simplified translation for it

gexgd0419:
Noticed that there's no translation item for the list header text "Sandbox" in SelectBoxWindow in Qt Linguist and it's always shown in English, then I found that it's non translatable, so I changed it and provided zh-CN translation for it.
If it's non translatable for a reason, close this request but I'd still like to know the reason.


isaak654:
It looks like even Linguist doesn't recognize it on the UI:
immagine

PR #1320, opened 2021-10-20
Make some strings translatable & update Chinese Simplified translation

gexgd0419:
Noticed that the "Support Sandboxie-Plus on Patreon" label and the title of the color dialog is not translatable.

They were made translatable with `tr()`.


isaak654:
Thank you for providing an overview about all the possible options.

PR #1337, opened 2021-10-26
Replace default hotkey with Shift+Pause

isaak654:
This is only related to Plus UI. If you set `Ctrl+Alt+Pause` hotkey on Options -> Global Settings -> General Config, this will be incorrectly displayed  as `Ctrl+Alt+Cancel`

Further reasons for changing the current hotkey (which terminates all boxed processes) were provided [here](https://github.com/sandboxie-plus/Sandboxie/pull/1337#issuecomment-952063294).

In a few words: the `Cancel` reference related to the default hotkey is automatically translated on the Plus UI as it was a Cancel button for dialogs. This conflict (partially due to how [Qt interpretes the key codes](https://github.com/sandboxie-plus/Sandboxie/pull/1337#issuecomment-952142434)) would be fixed by replacing the default `Ctrl+Alt+Cancel` with `Shift+Pause` or other alternatives I proposed [here](https://github.com/sandboxie-plus/Sandboxie/pull/1337#issuecomment-953220971).

### DIFF
--- a/SandboxiePlus/SandMan/SandMan.cpp
+++ b/SandboxiePlus/SandMan/SandMan.cpp
@@ -1024,7 +1024,7 @@ void CSandMan::SetupHotKeys()
 	m_pHotkeyManager->unregisterAllHotkeys();
 
 	if (theConf->GetBool("Options/EnablePanicKey", false))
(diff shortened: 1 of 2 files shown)


DavidXanatos:
Well I personally prefer ctrl + break, but its a problematic combination you cant even re enter it properly, strangely enough ctrl+scrol lock enters ctrl+cancel... so given that its so widely inconsistent i think shift + break is a good alternative indeed. So lets do that :)

PR #1371, opened 2021-11-14
Update FLT_FILE_NAME_INFORMATION usage to match docs

Therzok:
This is possible a performance issue I've noticed on my system, where the `FMfn tag` pool is using increasingly more memory.

It all seems be pointing to fltMgr.sys, which I guess is this `FltGetFileNameInformation` call.

The code-path seems to be querying `FltGetFileNameInformation`, and on the success path, it only released the filename when the spool folder work is allowed.

That meant this was effectively leaking a reference to `FLT_FILE_NAME_INFORMATION` for non-spooler work folders, because as the docs state:
> After a successful call to FltGetFileNameInformation, the caller is responsible for releasing the pointer returned in the FileNameInformation parameter when the pointer is no longer needed. The caller does this by calling FltReleaseFileNameInformation.

### DIFF
--- a/Sandboxie/core/drv/file_flt.c
+++ b/Sandboxie/core/drv/file_flt.c
@@ -414,8 +414,11 @@ _FX FLT_PREOP_CALLBACK_STATUS File_PreOperation(
 								Log_Msg_Process(MSG_1319, wcPid, (PWCHAR)pStr, proc->box->session_id, proc->pid);
                                 Mem_Free(pStr, len);
                             }
-                            FltReleaseFileNameInformation(pTargetFileNameInfo);
                         }   // if (FltGetFileNameInformation)
+
+                        if (pTargetFileNameInfo != NULL) {
+                            FltReleaseFileNameInformation(pTargetFileNameInfo);
+                        }
                     }   // if (proc)
                 }   // if (ulOwnerPid)
             }   // is this the print spooler process?


DavidXanatos:
great find will check it out on my system asap

Therzok:
Yeah, over 2 days of usage, I ended up with a lot of pools leaked - validated via poolmon. There were also a lot of mapped memory file handle leaks of files that have only been open temporarily, so I suspect this is what caused it.

DavidXanatos:
Yea there are many old bugs in the sandboxie code base.

isaak654:
It looks like this pull request may have stopped the ability to print inside sandboxes for some Chrome/Firefox users: https://github.com/sandboxie-plus/Sandboxie/discussions/1376#discussioncomment-1729695

DavidXanatos:
hmm... not good because we should avoid memory leaks still

isaak654:
Bug confirmed on 1.0.1 as well, just reproduced with sandboxed Mozilla Firefox and sandboxed Notepad++.
I'm going to revert this pull request if no change will be provided in the next days.

isaak654:
My bad, fortunately this PR isn't involved in the printing issue.

Therzok:
I was really confused how it would cause that, but did not know enough about the Sandboxie internals or FLT API to respond on that :D

DavidXanatos:
I really dont think its reaponsible, also when in roll back the change printing remains broken, so its somethign different

isaak654:
I can confirm version [1.0.2](https://github.com/sandboxie-plus/Sandboxie/releases/tag/1.0.2) fixed the printing issue.

PR #1383, opened 2021-11-18
Replace license shown in installer with correct Sandboxie license

CONIGUERO:
For some reason the license shown to the end-user at installation time is not the correct one. The corresponding license is the GPLv3 license under which both Sandboxie and Sandboxie plus are on.


DavidXanatos:
Only Sandboxie is GPLv3, Sandboxie Plus's SandMan.exe as it does not load any code from the old project and only uses regular IPC communication mechanisms has an own license, once that does not allow for commercial usage of sandman.exe without permission.

PR #1391, opened 2021-11-21
Add templates for SumatraPDF

Dyras:
Hello! This adds a template for SumatraPDF. Seems to work just fine using Sandboxie 1.0 and SumatraPDF 3.3.3.

I don't have a printer though, so if anyone wants to try if that works, that would be great.


isaak654:
> I don't have a printer though, so if anyone wants to try if that works, that would be great.

The test print works as expected with the ForceProcess setting applied to the SumatraPDF portable version.

Dyras:
Then this seems solid to me! *Knocks on wood*

Issue #1369, opened 2021-11-13
SetCursorPos sets wrong coordinates with high-DPI scaling screens


alvinhochun:
**Describe the bug**
I haven't checked thoroughly, but from what I've experienced with running Blender under Sandboxie, it wraps the mouse cursor around the viewport when performing transformations and viewport controls, and here it always wrap the cursor to the wrong position.

**To Reproduce**
1. Set monitor to 200% scaling
2. Run Blender under Sandboxie
3. Use the default file with the default cube
4. With the cursor inside the viewport area, press `g` to enter "Move" mode
5. Move the cursor past the edge of the viewport
6. Observe how the cursor moves. The cursor will be warped to the expected coordinates multiplied by two.

**Screenshots**

**Expected behavior**
The cursor should be wrapped to the exact opposite end of the viewport.

**System details and installed software**
 - Windows 10 Pro 20H1 64-bit
 - Sandboxie Plus 0.9.6 64-bit
 - I don't think this is a recent regression
 - Windows Defender is enabled (not that it would matter)

**If you have a compatibility issue**
 - Blender: https://www.blender.org/
 - Anything that calls `SetCursorPos` and is high-DPI aware

**Additional context**
Seeing that Sandboxie [hooks](https://github.com/sandboxie-plus/Sandboxie/blob/294966c7d6e99cd153ede87ad09aa39ef29e34c3/Sandboxie/core/dll/guimisc.c#L457) and [proxies](https://github.com/sandboxie-plus/Sandboxie/blob/294966c7d6e99cd153ede87ad09aa39ef29e34c3/Sandboxie/core/svc/GuiServer.cpp#L3334) the API call, I have a suspicion that it is due to the different DPI awareness of the Blender process and the Sandboxie GUI server (or whatever it is that actually performs the API call). Blender is per-monitor DPI aware (can be checked with Task Manager).

The `SendInput` API might also be affected by the same issue. There are also other WinAPIs which changes behaviour depending on the DPI awareness of the calling thread/process which might need attention.

DavidXanatos:
> Seeing that Sandboxie hooks and proxies the API call, I have a suspicion that it is due to the different DPI awareness of the Blender process and the Sandboxie GUI server (or whatever it is that actually performs the API call). Blender is per-monitor DPI aware (can be checked with Task Manager).
>

yes that is very likely the issue

alvinhochun:
[...]

PR #1394, opened 2021-11-22
Fix SetCursorPos and ClipCursor ignoring DPI awareness

alvinhochun:
This fixes the issue of `SetCursorPos` coordinates being scaled incorrectly when high-DPI display scaling is active. (#1369)

The `DPI_AWARENESS_CONTEXT` of the calling thread is passed to GuiServer, which is applied there before calling the actual WinAPI. The original `DPI_AWARENESS_CONTEXT` of the GuiServer thread is then restored immediately to prevent affecting other functions.

The same fix has been applied to `ClipCursor`, ~~but I do not have a way to test it~~. `SendInput` doesn't seem to be affect by the looks of it. I have not checked the other APIs.

### DIFF
--- a/Sandboxie/core/dll/guimisc.c
+++ b/Sandboxie/core/dll/guimisc.c
@@ -123,6 +123,11 @@ typedef int (*P_ReleaseDC)(
 
 static P_GetUserObjectInformationW __sys_GetUserObjectInformationW = NULL;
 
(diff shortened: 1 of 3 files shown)


alvinhochun:
Okay, I fixed these APIs for 32-bit processes, also verified with a simple WinForms program that `ClipCursor` was broken and fixed by this change.

PR #1430, opened 2021-12-12
Refresh and update French translations

Mmoi-Fr:
French translation update to the current state.
Little refactor of Text-French-1036.txt in order to be like the English one and be easilly diffable.

Note: it would be a good idea to do a cleanup in the translation files for all languages (there are a lot of commented texts and some maybe not used anymore texts ass well, i.e. 7933;txt;01)


isaak654:
@Mmoi-Fr It looks like the parsing script that handles Sbie messages files produced a conflict with your changes, so I had to revert Text-French-1036.txt in https://github.com/sandboxie-plus/Sandboxie/commit/cc45079dcfe1bbc19dc71c35de5ed196dfb3edf3

PR #1438, opened 2021-12-17
Refresh and update SimpChinese translations

Zerorigin:
1. Improve translation for Access Options: Closed, Open & Open for All.
2. Replace () with（）.
3. Update SimpChinese translation for most of the newly added strings.

### DIFF
--- a/SandboxiePlus/SandMan/sandman_zh_CN.ts
+++ b/SandboxiePlus/SandMan/sandman_zh_CN.ts
@@ -57,37 +57,37 @@
     <message>
         <location filename="Windows/NewBoxWindow.cpp" line="30"/>
         <source>New Box</source>
-        <translation>New Box</translation>
+        <translation>新沙盒</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


Zerorigin:
I did‘t translate the access option: Boxed Only.
Because I need to know more about the behavior of this option in order to provide a more relevant translation.

isaak654:
It looks like there is some error to fix on your translation: https://github.com/sandboxie-plus/Sandboxie/runs/4560097712?check_suite_focus=true

> lrelease error: Parse error at ..\SandMan\sandman_zh_CN.ts:3650:14: Opening and ending tag mismatch.

Zerorigin:
> It looks like there is an error to fix on your translation: https://github.com/sandboxie-plus/Sandboxie/runs/4560097712?check_suite_focus=true
> 
> > lrelease error: Parse error at ..\SandMan\sandman_zh_CN.ts:3650:14: Opening and ending tag mismatch.
> 
> missing_end_tag

yeh... I'm so sorry about miss the <.

PR #1453, opened 2021-12-22
Updated Languages.iss

mpheath:
All languages updated by use of a script to extract strings from the Classic msg txt files.

If the translators can confirm if any issues, that would be helpful.

The Portuguese is the brazilian type if I got that correct.


DavidXanatos:
thanks :)

PR #1457, opened 2021-12-23
Improve sorting of templates

isaak654:


isaak654:
@APMichael What do you think about these sorting changes to the _Templates.ini_?

APMichael:
> @APMichael What do you think about these sorting changes to the _Templates.ini_?

Thank you for the interest in my opinion. I really like the improved sorting, because everything becomes much clearer. 👍

Issue #1102, opened 2021-08-13
When I change VMWare Workstation Pro 16 (16.1.2 build-17966106) Virtual Network Editor in Sandboxie, the change affect the real system


0x391F:
**Describe the bug**
When I change VMWare Workstation Pro 16 (16.1.2 build-17966106) Virtual Network Editor in Sandboxie, the change affect out of the Sandboxie

**To Reproduce**
Original status:
![1](https://user-images.githubusercontent.com/56181919/129341092-2b7bb8af-a15f-467d-af2e-2fbca5bd38f9.png)

1. Run VMWare Workstation Pro in Sandboxie with UAC Administrator privilege (otherwise couldn't startup any VM in Sandboxie)
2. Change the bridge adapter from Auto to physical network adapter (Realtek PCIe GbE Family Controller). 
![2](https://user-images.githubusercontent.com/56181919/129341135-247a1921-a692-45cf-8515-61aa7a340134.png)

3. Click "OK" and exit
4. Clean up the Sandboxie
5. Run VMWare Workstation Pro, the bridge adapter changed to physical network adapter (Realtek PCIe GbE Family Controller)
![3](https://user-images.githubusercontent.com/56181919/129341153-d76b828d-aea9-496e-9f46-f7144e2559e4.png)

**Expected behavior**
Any change inside Sandboxie shouldn't affect out of it. 
**Screenshots**

**System details and installed software (please provide the following information):**
 - What is your Windows edition and version? Windows 10 Enterprise LTSC 2019 (1809).
 - What is your current Sandboxie edition and version? (Sandboxie Plus 0.9.3).

DavidXanatos:
I have tested this and to me it looks like all changes are contained in the sandbox i.e. you can change stuff in the editor but the changes are contained in the sandbox and not propagated outside

DavidXanatos:
That is strange, changing other settings is contained but changing the bind adapter seams not very wired

DavidXanatos:
So as far as I can tell the device "\Device\VMnetUserif" allows to configure VMNet0
its strange that its accessible its not configured open,
if its configured explicitly closed VMnet0 is no longer listed at all.

so as a workaround you can close that file path

isaak654:
> 
> so as a workaround you can close that file path

`ClosedFilePath=\Device\VMnetUserif`

isaak654:
How would you use a bridge network in Sandboxie? What's your use case?

[...]

PR #1481, opened 2021-12-27
[Reminder] Fix isolation flaw with IOCTLs on non-hardened sandboxes

0x391F:
I think compare with the change affect out of the Sandboxie, VMNet0(Bridging network) could't work is even worse.

---

Reply n. 1 on 3 Jan 2022: https://github.com/sandboxie-plus/Sandboxie/issues/1102#issuecomment-1004026992

> It is a general flaw in sbies security architecture to allow Device IO Control access to random devices at \Device* this issue will be remedied in a later release.
> for the issue at hand the result will be same as ClosedFilePath=\Device\VMnetUserif
> hence if one wants to access that device OpenFilePath=\Device\VMnetUserif will have to be added
> resulting in the initial issue at hand, device is open sandboxed apps can talk to it device driver modifies outside OS.
> 
> It is possible to add a driver level control code filter for this device, but as its not a part of windows and will be closed by default, its not something I would want to spend my time on,
> i add a fix for one device driver and the world stays full of millions other device drviers that a good portion of will also allow to compromise the system still.
> 
> If this is something extremely important for someone, them one can become a patron at one of the 3 highest tears, then I would look into that custom solution for this particular use case.

Reply n. 2 on 7 Apr 2022: https://github.com/sandboxie-plus/Sandboxie/issues/1102#issuecomment-1091668084

> In one of the future builds a new hardened box mode will be introduced that prevents access to not opened device endpoints that will be the fix.
> The user must choose if he wants to allow a particular sandbox access to this device or not. Its not reasonable to implement custom filtering for that endpoint.
> If that means that then a sand boxed process can alter that tools configuration so be it.
> Until then you can use ClosedFilePath=!vmware.exe,\Device\VMnetUserif or alike to allow only vmware.exe to access that endpoint but prevent all other processes from doing the same.

### DIFF
--- a/Sandboxie/install/Templates.ini
+++ b/Sandboxie/install/Templates.ini
@@ -167,10 +167,6 @@ ClosedClsid={4991D34B-80A1-4291-83B6-3328366B9097}
 ClosedClsid={C2F03A33-21F5-47FA-B4BB-156362A2F239}
 
 [Template_ThirdPartyIsolation]
-# close VMNet0 virtual network
-ClosedFilePath=\Device\VMnetUserif
-
-
 #
 # Internet Explorer
 #


DavidXanatos:
Containing a driver inside sandboxie, i was thinking about that at the bginning, given that Wine (this linux thingy) can load some drivers and show them a fake windows kernel,
but while somewhat doable, I believe this to be a huge waste of time, so I wont be dining that, unless someone wants hier me to implement exactly this feature.

Also in case of VMWare this would work but not function as the point is to have access to the real windows network stack.

Also since we now have device isolation 
grafik


I will close this.

PR #1497, opened 2022-01-02
Added Portuguese of Portugal

JNylson:


isaak654:
1) Your change implies that the existing `sandman_pt.ts` file should be renamed as `sandman_pt_BR.ts`.
2) There would be also multiple references to update:
https://github.com/sandboxie-plus/Sandboxie/search?q=sandman_pt
https://github.com/sandboxie-plus/Sandboxie/blob/54580a85480f1cdda805bb69709979a107ba7b35/Installer/Sandboxie-Plus.iss#L213
3) Are you going to maintain both TS files?

JNylson:
My native language is pt_BR, ideally someone was to translate it, but I can try for now

JNylson:
thanks for fixes

isaak654:
I think this should be planned for the next 1.0.8, if nothing else comes up.

mpheath (approved):
Changed a tab indent that should have been 4 spaces. The iss entries appear OK. I give it a :+1:

PR #1526, opened 2022-01-07
Update outdated VCRedist references

isaak654:

PR #1533, opened 2022-01-09
Minor classic installer fixes

sredna:
- No need for System::Call when there is a constant for it.
- Use InetC for downloads.
- Don't use root of %temp% to avoid conflicts and DLL hell.

### DIFF
--- a/Sandboxie/install/SandboxieVS.nsi
+++ b/Sandboxie/install/SandboxieVS.nsi
@@ -821,7 +821,7 @@ FunctionEnd
 
 Function DownloadStatPng
 ;
-;	Pop $0	; Get the parameter (file name to download)
+	Pop $0	; Get the parameter (file name to download)
 ;	${If} ${RunningX64}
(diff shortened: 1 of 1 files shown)


isaak654:
About the `DownloadInstallVCRedist` part, I just verified that it doesn't check if the DLLs are present, it just compares the DLLs when they're located on the system32 folder.

Right now I'm on a clean W10 that doesn't contain by default those DLLs or the ones I used [here](https://github.com/sandboxie-plus/Sandboxie/pull/1526), so this special case is not covered (according to my test) and the Sbie Installer won't tell you anything, but Classic still runs regardless of them.

So I would like to know your opinion about this behavior before merging these changes.

Thank you for your contribution in this repository!

sredna:
The "MS libraries are all static with Visual Studio 2015" comment implies that you don't need any of the C-runtime DLLs. If that is the case then all that stuff can remain unused.

PR #1543, opened 2022-01-14
fix string

hg421:
partially undo 1780e93b81986641005e26e0d9176778515014e5, "%1" is used here to conditionally insert the phrase ", but it remains valid for the current build"


isaak654:
Thanks!

PR #1571, opened 2022-01-26
Fix off-by-one vulnerability of listpids

7eRoM:
In `/listpids` command to show list of current active processes related to sandboxie, `n+1` would be read off the array.


DavidXanatos:
thanks good catch, that bug was repeated in a few more places, now all is fixed

PR #1572, opened 2022-01-26
Change Unix (LF) to Windows (CRLF) to pass Parse post-link script

7eRoM:
After compiling `Parse` project, the post-link script would be failed because the format of `Sbie-English-1033.txt` is in Unix (LF) format. I saved the file in Windows (CRLF) format.


isaak654:
I suggest to apply it for all files on the msgs folder.

7eRoM:
> I suggest to apply it for all files on the msgs folder.

All Done!

isaak654:
That happened because we struggled with a faulty .gitattributes, and a lot of files were changed to LF (in other folders too).
My only doubt now is about other script files on the msgs folder, do you think they could stay in LF?

7eRoM:
> 

Others are safe. I've tested them locally on my machine multiple times and got no errors. The only and main problem was related to these txt files cause the post-link script of `Parse` to be failed and compiling other projects as well.

isaak654 (approved):

PR #1592, opened 2022-02-04
Update sandman_de.ts

bastik-1001:
Translated new strings to German and update the changed strings.

### DIFF
--- a/SandboxiePlus/SandMan/sandman_de.ts
+++ b/SandboxiePlus/SandMan/sandman_de.ts
@@ -839,7 +839,7 @@ Vollständiger Pfad: %4</translation>
     <message>
         <location filename="Windows/RecoveryWindow.cpp" line="104"/>
         <source>Delete everything, including all snapshots</source>
-        <translation type="unfinished"></translation>
+        <translation>Alles löschen, inklusive aller Schnappschüsse</translation>
     </message>
     <message>
         <location filename="Windows/RecoveryWindow.cpp" line="128"/>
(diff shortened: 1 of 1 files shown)


isaak654:
Nice job! I have just a small hint for the future language updates (extended to all translators in this repository).

In the [Localization notes and tips](https://github.com/sandboxie-plus/Sandboxie/discussions/1123#discussioncomment-1203489) I provided two commands to sync manually the new language entries (introduced in each Plus version to the Qt source files) in all .ts files.

Currently there is no automatic task for this operation (I have no idea how it could be written), so I would suggest occasionally to run one of them in order to produce commits like [this](https://github.com/sandboxie-plus/Sandboxie/commit/c29842fc300c79ef38f7d64ce26a2e96852c50c9) for all languages, especially if there are no english typos to correct in the source files.

isaak654 (approved):

bastik-1001:
I had looked into that before and had decided to wait until the .ts files got updated by the maintainer, as I could not update the .ts file with lupdate or lupdate-pro, providing it with qmake, as it wanted Qt to be installed.

Qt Lingust works fine, but can't update the files itself. I shied away from creating an account and installing a full framework just to be able to update the .ts file, if I even succeed at using the tool.

If it is fine, I still stick to using Linguist to update my translation, as I still don't understand why the creators of Qt made the process of updating a language file rather complicated.

isaak654:
Please help to spread the word about this, thanks:
[ [Plus UI] Create a new workflow to improve delivery of language updates](https://github.com/sandboxie-plus/Sandboxie/issues/1595)

A minimal discussion and evaluation of ideas should be started, at least.

PR #1593, opened 2022-02-04
Update sandman_zh_CN.ts

BecodReyes:
Chinese translation follow up,
and translate "sandbox" uniformly as "沙盒".

### DIFF
--- a/SandboxiePlus/SandMan/sandman_zh_CN.ts
+++ b/SandboxiePlus/SandMan/sandman_zh_CN.ts
@@ -1701,7 +1701,7 @@ No will choose: %2</source>
         <location filename="SandMan.cpp" line="245"/>
         <location filename="SandMan.cpp" line="386"/>
         <source>Pause Forced Programs Rules</source>
-        <translation>禁用必沙程序</translation>
+        <translation>停用必沙程序</translation>
     </message>
     <message>
         <location filename="SandMan.cpp" line="940"/>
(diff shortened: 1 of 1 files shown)


SandboxiePlus/SandMan/sandman_zh_CN.ts
```diff
@@ -4397,7 +4397,7 @@ Please note that this values are currently user specific and saved globally for
         <location filename="Forms/SettingsWindow.ui" line="619"/>
         <source>Only Administrator user accounts can use Pause Forced Programs Rules command</source>
         <oldsource>Only Administrator user accounts can use Disable Forced Programs command</oldsource>
-        <translation type="unfinished">仅管理员账户可使用“禁用必沙程序”命令</translation>
+        <translation>仅管理员账户可使用“禁用必沙程序”命令</translation>
```
isaak654, line 4400: Sorry, but there is no translation here for "Pause Forced Programs Rules".

1) "Rules" word was not translated, please fix it. 
2) You may also consider to replace "Disable" with "Pause" in your language.

Please make sure to apply the suggestion(s) also in line 1704 and line 1724

isaak654 (changes requested):

SandboxiePlus/SandMan/sandman_zh_CN.ts
BecodReyes: Oh, that was due to an oversight on my part.
I have modified it now.
Thank you for pointing out the oversight I made.

Issue #1602, opened 2022-02-07
The new BlockSoftwareUpdaters template needs more browser support


isaak654:
### Is your feature request related to a problem or use case?

Only a couple of browsers are supported for the new introduced BlockSoftwareUpdaters template:

https://github.com/sandboxie-plus/Sandboxie/blob/2be7d6cb033041ec3ef2c4eebb3c7e6e46c90461/Sandboxie/install/Templates.ini#L162-L178

### Describe the solution you'd like

Add the rest of supported browsers that have already a template:

- [x] Waterfox
- [x] PaleMoon
- [x] SeaMonkey

~~LibreWolf~~ (no built-in updater)

- [x] MS Edge
- [x] Vivaldi
- [x] Brave
- [x] Opera
- [x] Yandex

~~Ungoogled~~ (no built-in updater)
~~Iron~~ (no built-in updater)

- [ ] Maxthon_6
- [ ] Comodo Dragon
- [ ] Osiris
- [ ] Slimjet

Dyras:
Hello!

I know from experience that at least Edge and Chrome don't try to update themselves in the sandbox. Maybe Brave too. But this is definitely needed for Waterfox, PaleMoon and SeaMonkey.

I can take a look at this later this week if you want.

isaak654:
Thank you for taking care about it!

Unfortunately it seems that other users had different experiences about the updates within the sandboxes, so I think all supported browsers would still need to be reconsidered:

> Well, Chrome gets updated automatically, even running within the Sandbox, but instead of a successful update gets crippled that way. You can only disable forced programs intentionally and beforehand if you know exactly when that update is going to be triggered.
Opera gets updated once a week, although not necessarily automatically. And Chrome more often than once a month, too, albeit constituting the bigger problem because of the unpredictable point in time such an upgrade is triggered. 

I'm also wondering... what about the Firefox/Chrome antiphishing templates enabled by default?

They are more needed or less needed, now that BlockSoftwareUpdaters is forcefully enabled for anybody?
And if they are needed, why MS Edge is out of the equation?

Since @APMichael took care of browser templates in the past, I'm going to attach it to the discussion.

isaak654:
Thank you for taking care about it!

Unfortunately it seems that other users had different experiences about the updates within the sandboxes, so I think all supported browsers would still need to be reconsidered:

[...]

PR #1630, opened 2022-02-14
Adds main browsers to BlockSoftwareUpdaters template

Dyras:
Hello. This adds 3 new browsers to BlockSoftwareUpdaters. Planning on throwing in Edge, Vivaldi and Pale Moon tomorrow.

The plan is to support all browsers mentioned here: https://github.com/sandboxie-plus/Sandboxie/issues/1602

### DIFF
--- a/Sandboxie/install/Templates.ini
+++ b/Sandboxie/install/Templates.ini
@@ -178,8 +178,10 @@ SoftwareUpdater=thunderbird.exe,*\mozilla*\updates\*updater.exe
 SoftwareUpdater=GoogleUpdate.exe,*\google\update\*chrome_installer.exe
 
 SoftwareUpdater=msedge.exe,*\Microsoft\EdgeUpdate\MicrosoftEdgeUpdate.exe
+SoftwareUpdater=msedge.exe,*\Microsoft\EdgeUpdate\*MicrosoftEdgeUpdateBroker.exe
 
 SoftwareUpdater=brave.exe,*\BraveSoftware\Update\BraveUpdate.exe
+SoftwareUpdater=brave.exe,*\BraveSoftware\Update\*BraveUpdateBroker.exe
 
 SoftwareUpdater=waterfox.exe,*\Waterfox\updater.exe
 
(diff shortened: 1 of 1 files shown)


Sandboxie/install/Templates.ini
```diff
@@ -177,6 +177,18 @@ SoftwareUpdater=thunderbird.exe,*\mozilla*\updates\*updater.exe
...
 
+SoftwareUpdater=msedge.exe,*\Microsoft\EdgeUpdate\MicrosoftEdgeUpdate.exe
+
+SoftwareUpdater=brave.exe,*\BraveSoftware\Update\BraveUpdate.exe
+
+SoftwareUpdater=waterfox.exe,*\Waterfox\updater.exe
+
+SoftwareUpdater=palemoon.exe,*\Pale Moon\updater.exe
+
+SoftwareUpdater=opera.exe,*\Opera\opera_autoupdate.exe
```
isaak654, line 188: It would be probably better to add below:
`SoftwareUpdater=launcher.exe,*\Opera\opera_autoupdate.exe`

Source: 
https://www.reddit.com/r/operabrowser/comments/goiniy/update_is_automatically_launched_by_exe_and_there

isaak654 (changes requested):

Sandboxie/install/Templates.ini
Dyras: Planning on adding all the browsers first, then I'll think about the sorting. Will probably be done early next week.

Dyras:
Every mainstream web browser is added now. I'm thinking I will do the minor ones next week or something. I've been busy with school and work lately

isaak654:
Before merging, do you have any suggestion about [this](https://github.com/sandboxie-plus/Sandboxie/issues/1602#issuecomment-1059333850)?

Dyras:
I think we could disable it and see how that will work out with people. If it causes confusion we can always consider adding it back in a changed form

(A helicopter has been hovering outside my apartment for the past 4 hours so if that makes no sense, that's why)

isaak654 (approved):
It looks good now, next time is the turn of minor browsers.

PR #1680, opened 2022-03-12
Fix SandMan Settings UI showing inverse of actual value for AutoRunSoftCompat

### LINKED ISSUE #1678: [Plus UI] AutoRunSoftCompat checkbox not working correctly
This is the checkbox in question:
compat_checkbox

There is no way to turn it on from the UI, it remains always off even after pressing OK button.

In order to reproduce it, you need to have `AutoRunSoftCompat=false` previously set on Sandboxie-Plus.ini.

If you replace it with `AutoRunSoftCompat=true`, the issue will be the opposite, that is, you can't turn off the checkbox from the UI at all.

mitchcapper:
Closes #1678

I am still in the multi-hour QT install so didn't actually test but this is basic enough.   in SandMan UI Settings the option variable was inverted for readability.  For comparability it is still stored the same way.

Setting the value from the ui does this correctly:
`theConf->SetValue("Options/AutoRunSoftCompat", !ui.chkNoCompat->isChecked());`
reading from the UI previously was not inverting the value:
`ui.chkNoCompat->setChecked(theConf->GetBool("Options/AutoRunSoftCompat", true));`

### DIFF
--- a/SandboxiePlus/SandMan/Windows/SettingsWindow.cpp
+++ b/SandboxiePlus/SandMan/Windows/SettingsWindow.cpp
@@ -153,7 +153,7 @@ CSettingsWindow::CSettingsWindow(QWidget *parent)
 	connect(ui.btnDelCompat, SIGNAL(clicked(bool)), this, SLOT(OnDelCompat()));
 	m_CompatLoaded = 0;
 	m_CompatChanged = false;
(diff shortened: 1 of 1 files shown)


DavidXanatos:
thanks looks good

PR #1689, opened 2022-03-14
Update French translation

hippalectryon-0:
Updated French translation, up to date with 477fc52691c5213348a456cf15b6ec90efaa1941:
* ran `lupdate`
* translated all the new and updated strings in `sandman_fr.ts`

[edit: solved] CONFIRMATION NEEDED: What's the difference between "Forced Programs" and "Forcing Programs" ? (ex in "Pause Forcing Programs command") In doubt, I treated those as the same in the translation so far...

### DIFF
--- a/SandboxiePlus/SandMan/sandman_fr.ts
+++ b/SandboxiePlus/SandMan/sandman_fr.ts
@@ -1493,8 +1493,7 @@ Non choisira&#xa0;: %2</translation>
     <message>
         <location filename="SandMan.cpp" line="2582"/>
         <source>The supporter certificate has expired%1, please get an updated certificate</source>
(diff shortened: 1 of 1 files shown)


isaak654:
> CONFIRMATION NEEDED: What's the difference between "Forced Programs" and "Forcing Programs" ? (ex in "Pause Forcing Programs command") In doubt, I treated those as the same in the translation so far...

I think the second term is something like "Suspendre le forçage des programmes", but in my language I used the translation for "Pause Forced Programs" because it's shorter.

hippalectryon-0:
All good on my end then :)

isaak654 (approved):

PR #1726, opened 2022-03-25
Change license.txt and fix "New Box" translation

totsukasaika:
## Change license.txt
Sandboxie is provided under the GPLv3, modifications must be released under the same license when distributing the licensed material.

please see: https://choosealicense.com/licenses/gpl-3.0/

## fix "New Box" translation
The sandbox name can contain only letters, digits and underscores.


isaak654:
The GPLv3 license is already mentioned in these files: 
* [LICENSE](https://github.com/sandboxie-plus/Sandboxie/blob/master/LICENSE) Sandboxie
* [install/LICENSE.TXT](https://github.com/sandboxie-plus/Sandboxie/blob/master/Sandboxie/install/LICENSE.TXT) - Sandboxie Classic Installer

While [Installer/license.txt](https://github.com/sandboxie-plus/Sandboxie/pull/1726/files#diff-511832906cfe39a6d264bdcc35404f894ed643dc37a3fbc43f18f132979728cd) seems more related to the Sandman installer that is not under the GPLv3: https://github.com/sandboxie-plus/Sandboxie/pull/1383#issuecomment-974326425

> Only Sandboxie is GPLv3, Sandboxie Plus's SandMan.exe as it does not load any code from the old project and only uses regular IPC communication mechanisms hence it has an own license, once that does not allow for commercial usage of SandMan.exe without permission. The components are distributed as a aggregation and if you want 100% GPL you can always stick with using SbieCtrl.exe

Suggestion: You may want to edit [Installer/README.md](https://github.com/sandboxie-plus/Sandboxie/pull/1726/files#diff-6ba86e3f6d7c8cbf2f7a16fc6978e3a2a73e3c6ec7648394e1da374ec8e47a05) in order to write how to create Sandboxie-Plus installers, as it was done for Sbie Classic: [install/ReadMe.txt](https://github.com/sandboxie-plus/Sandboxie/blob/master/Sandboxie/install/ReadMe.txt).

isaak654 (changes requested):
The GPLv3 license is already mentioned in these files: 
* [LICENSE](https://github.com/sandboxie-plus/Sandboxie/blob/master/LICENSE) Sandboxie
* [install/LICENSE.TXT](https://github.com/sandboxie-plus/Sandboxie/blob/master/Sandboxie/install/LICENSE.TXT) - Sandboxie Classic Installer

While [Installer/license.txt](https://github.com/sandboxie-plus/Sandboxie/blob/master/Installer/license.txt) seems more related to the Sandman installer that is not under the GPLv3: https://github.com/sandboxie-plus/Sandboxie/pull/1383#issuecomment-974326425

> Only Sandboxie is GPLv3, Sandboxie Plus's SandMan.exe as it does not load any code from the old project and only uses regular IPC communication mechanisms hence it has an own license, once that does not allow for commercial usage of SandMan.exe without permission. The components are distributed as a aggregation and if you want 100% GPL you can always stick with using SbieCtrl.exe

Suggestion: You may want to edit [Installer/README.md](https://github.com/sandboxie-plus/Sandboxie/pull/1726/files#diff-6ba86e3f6d7c8cbf2f7a16fc6978e3a2a73e3c6ec7648394e1da374ec8e47a05) in order to write how to create Sandboxie-Plus installers, as it was done for Sbie Classic: [install/ReadMe.txt](https://github.com/sandboxie-plus/Sandboxie/blob/master/Sandboxie/install/ReadMe.txt).

SandboxiePlus/SandMan/sandman_zh_TW.ts
```diff
@@ -65,7 +65,8 @@
     <message>
         <location filename="Windows/NewBoxWindow.cpp" line="30"/>
         <source>New Box</source>
-        <translation>新沙箱</translation>
+        <translatorcomment>沙箱名稱只能包含字母、數字和下劃線，不應對此處的文本進行翻譯！</translatorcomment>
+        <translation>New Box</translation>
```
isaak654, line 69: This needs to be ported on a dedicated PR.

Also, I would suggest to inform the most active translator @Hulen, since it looks like the mentioned string has been changed a number of times already.

isaak654 (changes requested):

DavidXanatos:
@totsukasaika the license of the new binaries which were written from scratch and that run in own processes will not be changed, aggregation of GPL tools with other non GPL tools is a common industry praxis.

SandboxiePlus/SandMan/sandman_zh_TW.ts
Hulen: Ok, I'll pay attention to this later. Thank you.

PR #1730, opened 2022-03-26
Update sandman_zh_TW.ts

Hulen:
fix "New Box" translation.
thanks @totsukasaika .

PR #1731, opened 2022-03-26
Update sandman_zh_TW.ts v2

Hulen:
new strings translation & fix "New Box" translation.


isaak654:
#1730

isaak654:
I doubt there is a method to prevent the future appearance of character entities like `&quot;` or `&apos;`
They will most likely come back next time.

isaak654 (approved):

PR #1770, opened 2022-04-06
fix windows credentials

weihongx9315:
Completed the credential implementation of the sandbox, after the fix, the password can be saved correctly when the rdp is connected


DavidXanatos (commented):
just removing the error messages from the A versions is wrong as they are not properly redirected as the w versions is the tools for which you fixed this using the A versions?
I think first we need to find out why sandboxie redirects CredWrite to use WriteItem instead and then add a proper implementation for the A functions. Can you please privide a couple test cases so i can test the fix in praxis?

weihongx9315:
> just removing the error messages from the A versions is wrong as they are not properly redirected as the w versions is the tools for which you fixed this using the A versions? I think first we need to find out why sandboxie redirects CredWrite to use WriteItem instead and then add a proper implementation for the A functions. Can you please privide a couple test cases so i can test the fix in praxis?

CredWriteA/W is not the same as other win32 api(such as getaddrinfo/GetAddrInfoW),other A api internal calls W api,but CredWriteA not.Sbie redirects this api's purpose is not for 'redirect',it implements a whole  of credential methods,so it just use WriteItem to do it.But the question is,when it saves credential username and password,it writes wrong password...
To test this is very simple,you can use mstsc.exe to connect another windows PC and choose save password,you can see  it doesn't work~

APMichael:
I am not sure if this information is helpful for this:

"... Note: Sandboxie stores credentials in the sandboxed protected storage ..."
https://sandboxie-plus.com/sandboxie/opencredentials/

"... The Sandboxie PStore is stored in the file SbiePst.dat in the sandboxed Windows folder ..."
https://sandboxie-plus.com/sandboxie/protectedstorage/

weihongx9315:
> I am not sure if this information is helpful for this:
> 
> "... Note: Sandboxie stores credentials in the sandboxed protected storage ..." https://sandboxie-plus.com/sandboxie/opencredentials/
> 
> "... The Sandboxie PStore is stored in the file SbiePst.dat in the sandboxed Windows folder ..." https://sandboxie-plus.com/sandboxie/protectedstorage/

sbie implement the credentials itself,but it doesn't work as normal .That's why I fixed it,before this commit , you can test mstsc.exe to save password

DavidXanatos:
@weihongx9315  this is still an open issue but it requriers as said a more in depth analysis and a propper fix, to just get it to work you can use OpenCredentials=y

What sandboxie does is quite more elaborate, it changes a cred entry which would be saved by windows globally into a sort that gets saved to a file a file which is then sandboxed, so the saved credentials are contained within the sandbox.
It needs to be investigated why this mechanism no longer works and how it can be repaired, as we don't want things to be saved globally by default.

DavidXanatos:
A workable fix will be included in the next build, the issue was with Cred_GetName, it seams that its not correct to save the type as part of the name, for now i just disabled adding type to the name string, but presumably properly the fix should be applied when names are compared and allow 1 as wildcard for any when reading credentials.

weihongx9315:
> A workable fix will be included in the next build v1.1.1, the issue was with Cred_GetName, it seems that it's not correct to save the type as part of the name, for now i just disabled adding type to the name string, but presumably the fix should be properly applied when names are compared and allow 1 as wildcard for any when reading credentials.

Yes,the real reason is Cred_GetName，and my approach is to modify this, so that the credentials will not leak out of the sandbox. This function just  write the wrong credentials，so we just need to fix this

Issue #1716, opened 2022-03-23
Cannot start explorer


inkuxuan:
### What happened?

Starting explorer using "Run" -  "Boxed Tools" - "Windows Explorer" in the right-click menu at a sandbox will end up with Start.exe and explorer.exe being started and instantly terminated (No window pops up).
![image](https://user-images.githubusercontent.com/19793388/159733131-f5138dfc-947a-4b80-8e44-2b26ba15f7c8.png)

Starting explorer using "Run" - "Run Program" and entering "explorer.exe" will end up with multiple errors.
![image](https://user-images.githubusercontent.com/19793388/159733329-002744ef-0b79-41b5-aa37-dff3452fec9c.png)
![image](https://user-images.githubusercontent.com/19793388/159733413-3b42cf00-10e3-40f3-b209-0d00079a2607.png)

This problem is not present in Sandboxie-Plus v0.9.x (It occurs after I updated it)

### Download link

Not relevant

### To Reproduce

1. Create a standard sandbox
2. In "Sandbox Options" - "Resource Access", add an IPC Path for "All Program" with "Open" access for "*" (You have to enter this for reproducing this bug)
3. Save options and start explorer in the created sandbox

### Expected behavior

File Explorer starts

### What is your Windows edition and version?

Windows 10 Pro 20H2

### In which Windows account you have this problem?

A local or Microsoft account without special changes.

### Please mention any installed security software

Huorong Security 5.0.66.3

### What version of Sandboxie are you running?

Sandboxie-Plus v1.0.10

### In which sandbox type you have this problem?

In a Standard isolation sandbox (yellow sandbox icon).

### Where is the program located?

Not relevant to my request.

### Can you reproduce this problem on an empty sandbox?

I can confirm it also on an empty sandbox.

isaak654:
> Starting explorer using "Run" - "Boxed Tools" - "Windows Explorer" in the right-click menu at a sandbox will end up with Start.exe and explorer.exe being started and instantly terminated (No window pops up).

> This problem is not present in Sandboxie-Plus v0.9.x (It occurs after I updated it)

My downgrades confirmed that `OpenIpcPath=*` crashes explorer.exe at least since [0.9.8 x64](https://github.com/sandboxie-plus/Sandboxie/releases/tag/0.9.8)
Could you be more specific about which version introduced the issue, doing the rest of the downgrades version by version?

[...]

PR #1817, opened 2022-04-22
[help needed] Chinese Trans for SBIE2195

nkh0472:
I've noticed the msg SBIE2195 in `/Sandboxie/msgs/Sbie-English-1033.txt` comes from issue #1716.

https://github.com/nkh0472/Sandboxie/blob/e153bd91707d645908fdd6a23101f242aca70e9f/Sandboxie/msgs/Sbie-English-1033.txt#L382-L385

However, I can't fully understand this msg. Does it mean you have to remove any `COM object` rule in the `Resource Access` or something else?

Guess we have to make this msg clearer. Could be something like `SBIE2195 To run Explorer.exe sandboxed, the access rules for COM objects must not be Open.`


isaak654:
I think there are different ways to open COM, like this option:
Plus_Com_infrastructure

It writes the following lines:
````
OpenIpcPath=*\BaseNamedObjects*\__ComCatalogCache__
OpenIpcPath=\RPC Control\OLE*
OpenIpcPath=\RPC Control\LRPC*
OpenIpcPath=\RPC Control\epmapper
````

Another replacement would be `OpenIpcPath=*`, as far as I know it can only be typed manually.

The "COM infrastructure" definition was also mentioned in [v1.0.0 changelog](https://github.com/sandboxie-plus/Sandboxie/blob/090e000fae9466d45251faf121845753e66fe5fc/CHANGELOG.md#added-16), so maybe you could refer to it in this way:
`SBIE2195 To run Explorer.exe sandboxed, the access for COM infrastructure must not be Open.`

I currently have no idea how it looks on the UI, because I'm still on v1.0.15.

nkh0472:
> The "COM infrastructure" definition was also mentioned in [v1.0.0 changelog](https://github.com/sandboxie-plus/Sandboxie/blob/090e000fae9466d45251faf121845753e66fe5fc/CHANGELOG.md#added-16), so maybe you could refer to it in this way: `SBIE2195 To run Explorer.exe sandboxed, the access for COM infrastructure must not be Open.`

This is good enough for me. Thank you for your patience :)

PR #1874, opened 2022-05-20
fix:office ui

weihongx9315:
This commit fix bug of:us office 2016+ version (include 2019),you can see WINWORD.exe process is started,but you can't see the UI of WINWORD.exe.
The reason is when a process is started, but not the main thread load user32.dll,then other threads can't switch to the desktop,so hook ClientStartUp,to make sure every thread can switch to desktop.

### DIFF
--- a/Sandboxie/core/dll/gui.c
+++ b/Sandboxie/core/dll/gui.c
@@ -207,6 +207,10 @@ BOOLEAN Gui_UseProxyService = TRUE;
 
         P_GetOpenFileNameW          __sys_GetOpenFileNameW          = NULL;
 
+		typedef BOOL(*P_ClientThreadSetup)();
+		P_ClientThreadSetup       __sys_ClientThreadSetup = NULL;
+
+		static HMODULE            g_User32_Module = NULL;
 
 //---------------------------------------------------------------------------
 // Functions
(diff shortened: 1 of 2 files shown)


Sandboxie/core/dll/gui.c
```diff
@@ -854,7 +864,12 @@ _FX BOOLEAN Gui_ConnectToWindowStationAndDesktop(HMODULE User32)
     {
         return FALSE;
     }
+	if (TlsData->gui_connecting_desktop) {
+		// 避免重入
```
isaak654, line 868:
```suggestion
		// avoid re-entry
```

isaak654 (changes requested):

DavidXanatos:
I'm having issues running office in a normal box at all, but this fix is clearly for a normal box.
Do you install office with the click to run or the setup method?

DavidXanatos:
I have resolved my office issues mostly, nut cant reproduce the issue this fix fixes, for me even without it word starts up just fine now.
The fix looks safe enough and well made so we can pull it just in case it should not break anything, but i would prefer to first reproduce the issue. Currently running office 2021.

DavidXanatos:
I have investigated the issue a bit more and I wonder how the problem can happen in the first place, see https://github.com/sandboxie-plus/Sandboxie/blob/d93d4894aa503e1e3e83966524283116f7570fdf/Sandboxie/core/dll/dllmain.c#L148
the SbieDll.dll is calling Gui_ConnectToWindowStationAndDesktop already for each thread that is created.
Can you provide a call stack for ClientThreadSetup where the desktop initialization was missed?

Also the check for re-entry should be after the ConnectThread label IMHO

weihongx9315:
I use office 2019 with the setup method,in this case, the gui can't show. Gui_ConnectToWindowStationAndDesktop(NULL);  can't work, you can see if (!User32) return FALSE.

PR #1958, opened 2022-06-16
Update Turkish translation

fmbxnary:


isaak654 (changes requested):
In order to complete the translation:

1. You need to translate [these entries](https://github.com/fmbxnary/Sandboxie/blob/976004d1777a8a6855770371df7ab9fcc1aee905/Installer/Languages.iss#L42-L62)  and add them after the Spanish entries.
2. Uncomment [this line](https://github.com/fmbxnary/Sandboxie/blob/976004d1777a8a6855770371df7ab9fcc1aee905/Installer/Languages.iss#L23).
3. Uncomment [this line](https://github.com/sandboxie-plus/Sandboxie/blob/15b4b1e19dd63ecd91ac169bdddf7e8752deb16c/SandboxiePlus/SandMan/SandMan.pri#L83) too and arrange it in alphabetical order.

PR #2000, opened 2022-06-28
Add lupdate workflow

isaak654:


isaak654:
@mpheath I have a question for you, do you think it's achievable to create a script (to be included in a manual workflow) that could comment out `7901;txt;01` + the two lines below from all the [`Text-*.*.txt` files](https://github.com/sandboxie-plus/Sandboxie/tree/master/Sandboxie/msgs) and, if possible, extending the same pattern to other specified text IDs like 7936?

```
7901;txt;01
Sandboxie licence Manager
.
```

PR #2034, opened 2022-07-16
Minor string changes

Dyras:
Hello! This fixes a minor spelling error under the Sandboxie-Plus settings.

Edit: Perhaps it would be better to not use the word occupied at all. Maybe "used" would be better?

### DIFF
--- a/SandboxiePlus/SandMan/Forms/SettingsWindow.ui
+++ b/SandboxiePlus/SandMan/Forms/SettingsWindow.ui
@@ -57,7 +57,7 @@
            <item row="8" column="1" colspan="2">
             <widget class="QCheckBox" name="chkMonitorSize">
              <property name="text">
-              <string>Count and display the disk space ocupied by each sandbox</string>
+              <string>Count and display the disk space occupied by each sandbox</string>
              </property>
             </widget>
            </item>
--- a/SandboxiePlus/SandMan/Forms/SettingsWindow.ui
+++ b/SandboxiePlus/SandMan/Forms/SettingsWindow.ui
@@ -409,7 +409,7 @@
            <item row="10" column="2" colspan="2">
             <widget class="QLabel" name="label_11">
              <property name="text">
-              <string>* results may vary depending on the view mode</string>
+              <string>* a partially checked checkbox will leave the behavior to be determined by the view mode.</string>
              </property>
             </widget>
            </item>
--- a/SandboxiePlus/SandMan/Forms/SettingsWindow.ui
+++ b/SandboxiePlus/SandMan/Forms/SettingsWindow.ui
@@ -409,7 +409,7 @@
            <item row="10" column="2" colspan="2">
             <widget class="QLabel" name="label_11">
              <property name="text">
-              <string>* indetermined means depending on the selected view mode</string>
+              <string>* results may vary depending on the view mode</string>
              </property>
             </widget>
            </item>
--- a/SandboxiePlus/SandMan/Forms/SettingsWindow.ui
+++ b/SandboxiePlus/SandMan/Forms/SettingsWindow.ui
@@ -409,7 +409,7 @@
            <item row="10" column="2" colspan="2">
             <widget class="QLabel" name="label_11">
              <property name="text">
-              <string>* indetermined means depanding on the view mode</string>
+              <string>* indetermined means depending on the selected view mode</string>
              </property>
             </widget>
            </item>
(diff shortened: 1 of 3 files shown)


isaak654:
I think occupied is acceptable.

Next one: https://github.com/sandboxie-plus/Sandboxie/blob/58348273489a0c60aaf6a5c9404e613034d1b265/SandboxiePlus/SandMan/Forms/SettingsWindow.ui#L412

I think it should be translated as `* depending on the selected view mode`

If you fix this too, I'll make sure to update the pull request title.

SandboxiePlus/SandMan/Forms/SettingsWindow.ui
```diff
@@ -409,7 +409,7 @@
            <item row="10" column="2" colspan="2">
             <widget class="QLabel" name="label_11">
              <property name="text">
-              <string>* indetermined means depanding on the view mode</string>
+              <string>* indetermined means depending on the selected view mode</string>
```
isaak654, line 412:
```suggestion
              <string>* depending on the selected view mode</string>
```
Imho specifying also `indetermined means` can be confusing

isaak654 (changes requested):

SandboxiePlus/SandMan/sandman_en.ts
```diff
@@ -4897,7 +4897,7 @@ Please note that this values are currently user specific and saved globally for
     </message>
     <message>
         <location filename="Forms/SettingsWindow.ui" line="412"/>
-        <source>* indetermined means depanding on the view mode</source>
+        <source>* indetermined means depending on the selected view mode</source>
```
isaak654, line 4900: https://github.com/sandboxie-plus/Sandboxie/pull/2034/files#r922714887

isaak654 (changes requested):

SandboxiePlus/SandMan/Forms/SettingsWindow.ui
Dyras: "indetermined depending on the selected view mode" Sounds better. Should I change it to that?

isaak654: Mmmm, not quite. 🤔

Dyras: Where is this in the program anyway?

isaak654: The second one looks fine.

DavidXanatos:
<source>* indetermined means depending on the selected view mode</source>
        <source>* results may vary depending on the view mode</source>

is not quite right, this text must somehow indicate that the view mode only maters when the checkbox is in the intermediate/undetermined/half checked state

Dyras:
Then I'd suggest we change the string to include something that easier for the end user to understand. Something like:
* a gray box means the results will depend on the view mode

DavidXanatos:
well gray or something else depends on the ui style
best may be

"partially checked"

isaak654:
Alternatively:

*  `* partially checked options depend on the view mode`
* `* a partially checked box always affects the view mode`
* `* a partially checked box can affect the view mode`
* `* a partially checked box will affect the view mode`
* `* partially checked options will affect the view mode`

Dyras:
I think this one sounds the best
`* a partially checked box can affect the view mode`
The only question is if it should be "can" or "will". @DavidXanatos will partially checked boxes always affect the view mode, or only in some scenarios?

isaak654:
@Dyras 
David_answer
which means: 
`* a partially checked box will be affected by the view mode`

Dyras:
In my opinion
` * a partially checked checkbox will be affected by the view mode`
still sounds pretty confusing. It sounds to me as if _the checkbox_ will be affected by the view mode, not that the checkbox will have a different effect depending on the checkbox. 

Or maybe I'm just tired?

DavidXanatos:
Yea its not perfect I agree how about
* partially checked checkbox will leave the behavior to be determined by the view mode.

Dyras:
Sounds good to me!

DavidXanatos:
then please add it to the pull request and it can be merged

isaak654:
I would get rid of the changes made in sandman_en.ts, I already set a scheduled workflow that runs when needed: https://github.com/sandboxie-plus/Sandboxie/commit/1de13ea51e4faecbb0675e75c8eb901103927e57

isaak654:
I would get rid of the changes made in sandman_en.ts, now there is a scheduled workflow that runs when needed: https://github.com/sandboxie-plus/Sandboxie/commit/1de13ea51e4faecbb0675e75c8eb901103927e57

isaak654 (approved):

PR #2061, opened 2022-07-26
Changelog fixes

NewKidOnTheBlock:
- made sure there's a uniform 3 lines of space between each version


NewKidOnTheBlock:
I had problems understanding the first entry: `the access view list now adds in display the tailing * to file and key paths the same way the driver does ` -> I couldn't grasp what that means.

isaak654:
I agree, this time the explanation is very unclear and I struggle to find the change in the source code.

PR #2063, opened 2022-07-26
Update sandman_de.ts

bastik-1001:
Fresh fork to fix a spelling mistake that had been there for quite some time. Since I spotted it, I wanted to fix it right away, before I forget.


bastik-1001:
I updated the German Text file as well, since it was missing a string.

PR #2071, opened 2022-07-28
Update sandman_de.ts

bastik-1001:
Fresh fork to update the strings that got changed due to inconsistencies. Just so that it might make it into 1.2.7 before it gets released.


bastik-1001:
Just spotted and fixed another typo in my translation.

PR #2122, opened 2022-08-12
Update sandman_de.ts

bastik-1001:
Fresh fork to address new and old strings.

### DIFF
--- a/SandboxiePlus/SandMan/sandman_de.ts
+++ b/SandboxiePlus/SandMan/sandman_de.ts
@@ -3444,7 +3444,7 @@ Notiz: Die Updateprüfung ist oft zeitversetzt zu den letzten GitHub-Veröffentl
     <message>
         <location filename="Windows/SettingsWindow.cpp" line="881"/>
         <source>&lt;a href=&quot;check&quot;&gt;Check Now&lt;/a&gt;</source>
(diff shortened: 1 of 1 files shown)


SandboxiePlus/SandMan/sandman_de.ts
```diff
@@ -3479,22 +3479,22 @@ Notiz: Die Updateprüfung ist oft zeitversetzt zu den letzten GitHub-Veröffentl
     <message>
         <location filename="Windows/SettingsWindow.cpp" line="1095"/>
         <source>%1 (Current)</source>
-        <translation type="unfinished"></translation>
+        <translation type>%1 (Aktuell)</translation>
```
isaak654, line 3482:
```suggestion
        <translation>%1 (Aktuell)</translation>
```

isaak654 (changes requested):

bastik-1001:
My intention was to resolve the issue, by clicking on 'Resolve', but that did not what I expected, instead it got marked as resolved, so I fixed my file. There was a spelling mistake I fixed on top. Luckily, all checks have passed now.

I definitely made an error, and now that I saw it, I expected the build process to throw an error about a misformated file, not that the whole build process was going to fail.

PR #2142, opened 2022-08-15
Update Text-German-1031.txt

bastik-1001:
Fresh fork to provide some minor changes to Text-German-1031.txt

I looked for errors, but since most of the text was not written by me, there was less chance for me to f*ck it up. I even spelled "Sphäre" correct twice, even as I managed to screw it up for the .ts file.

### DIFF
--- a/Sandboxie/msgs/Text-German-1031.txt
+++ b/Sandboxie/msgs/Text-German-1031.txt
@@ -13,7 +13,7 @@ SBIE1102 Sandboxie-Treiber (SbieDrv) wird entladen
 .
 
 1103;evt;err;01
-SBIE1103 Sandboxie-Treiber (SbieDrv) version %2 konnte nicht gestartet werden
+SBIE1103 Sandboxie-Treiber (SbieDrv) Version %2 konnte nicht gestartet werden
 .
 
 1104;evt;err;01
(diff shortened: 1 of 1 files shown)


Sandboxie/msgs/Text-German-1031.txt
```diff
@@ -13,7 +13,7 @@ SBIE1102 Sandboxie-Treiber (SbieDrv) wird entladen
 .
 
 1103;evt;err;01
-SBIE1103 Sandboxie-Treiber (SbieDrv) version %2 konnte nicht gestartet werden
+SBIE1103 Sandboxie-Treiber (SbieDrv) Version %2 konnte nicht gestartet werden
```
isaak654, line 16: Is there a reason to change it? In English it is in lower case:

https://github.com/sandboxie-plus/Sandboxie/blob/6166ac0253f67bee141111e244ed33afe43dc3d3/Sandboxie/msgs/Sbie-English-1033.txt#L15-L17

bastik-1001: It's a noun, and those are spelled with an upper case letter in German. So to be correct, I changed the translation from a lower case letter to an upper case one.

PR #2173, opened 2022-08-21
Korean.isl - Update encoding

VenusGirl:
Changing the encoding because there was a problem with Korean not being displayed properly when installation

샌드박스 설치


isaak654:
Fair enough, but the line ending should be the same of the other .isl files (LF).

isaak654:
I think this should be properly tested by @DavidXanatos before merging, otherwise we risk running in circles.

DavidXanatos:
if this looks right: 
grafik


then the PR can be merged

VenusGirl:
Now I can see the Korean language correctly.

DavidXanatos:
I just saw the current 1.3.2 installer is missing the fix, a new one will be provided soon

DavidXanatos:
fixed installer uploaded please test

VenusGirl:
sandInstall01
sandinstall02

All are now well displayed when you install it.

PR #2191, opened 2022-08-29
Update sandman_de.ts

APMichael:
- Fixed strings with broken HTML.
- "Always on top" is now more usually translated.
- Translation for "Trace" aligned.


bastik-1001:
Thank you for the fixes and making the translation more consistent, as well as easier to understand.

PR #2192, opened 2022-08-29
Update sandman_de.ts

bastik-1001:
Fresh fork for the new strings translated.

### DIFF
--- a/SandboxiePlus/SandMan/sandman_de.ts
+++ b/SandboxiePlus/SandMan/sandman_de.ts
@@ -774,12 +774,12 @@ Notiz: Die Updateprüfung ist oft zeitversetzt zu den letzten GitHub-Veröffentl
     <message>
         <location filename="Windows/OptionsAccess.cpp" line="447"/>
         <source>Opening all IPC access also opens COM access, do you still want to restrict COM to the sandbox?</source>
-        <translation type="unfinished"></translation>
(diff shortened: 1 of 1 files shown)


SandboxiePlus/SandMan/sandman_de.ts
```diff
@@ -1256,7 +1256,7 @@ Vollständiger Pfad: %4</translation>
     <message>
         <location filename="SbiePlusAPI.cpp" line="310"/>
         <source>OPEN Root Access</source>
-        <translation type="unfinished"></translation>
+        <translation>ÖFFNE Rootzugriff</translation>
```
isaak654, line 1259: Deepl & Bing translators seem to suggest `Root-Zugriff ÖFFNEN`, so I was wondering how come.

isaak654 (changes requested):

SandboxiePlus/SandMan/sandman_de.ts
bastik-1001: Both are valid, I changed it to the suggested version.

I had chosen to keep the word that was emphasized at the start of the line.

As the services suggested, I added, the `-` to allow for easier reading. It could have been written as one word, but since it isn't a German word, it seems to be better to have it stand out.

DavidXanatos: it should be rather OFFENER Rootzugriff
as it does not say open as if in opening something but open as in it is open

bastik-1001: I changed it to be `OFFENER Rootzugriff`, as you said. Now I opted to make it one word, as it would look strange with the `-`

isaak654 (approved):

PR #2211, opened 2022-09-02
Add verbose messages

1mm0rt41PC:
Add SID of user used to run a process

### DIFF
--- a/Sandboxie/core/drv/process_util.c
+++ b/Sandboxie/core/drv/process_util.c
@@ -1205,8 +1205,28 @@ _FX void Process_LogMessage(PROCESS *proc, ULONG msgid)
 {
     BOX *box = proc->box;
     ULONG len = proc->image_name_len + box->name_len + 8 * sizeof(WCHAR);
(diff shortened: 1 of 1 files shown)


DavidXanatos:
This change will not be sufficient, as it will only affect driver messages, most messages are issued by the SbieDll.dll using the API_LOG_MESSAGE code invoking Api_LogMessage -> Log_Popup_MsgEx -> Api_AddMessage

Process_LogMessage which you have edited works like Process_LogMessage -> Log_Msg_Process -> Log_Popup_Msg -> Log_Popup_MsgEx -> Api_AddMessage

so the right place would be Log_Popup_MsgEx for example

or may be even in DriverAssist::LogMessage as here we have the pid of the process we could resolve the user to it and put that into the log file and only into the log file

DavidXanatos:
I have added the user name based on process ID to the service for buidl 1.3.3, log version will be 3:
https://github.com/sandboxie-plus/Sandboxie/commit/38b79b6d948bfe887209f94fb5c09f2d3577124c

PR #2215, opened 2022-09-05
Update sandman_tr.ts

fmbxnary:


SandboxiePlus/SandMan/sandman_tr.ts
```diff
@@ -1198,7 +1198,7 @@ Tam yol: %4</translation>
     <message>
         <location filename="SbiePlusAPI.cpp" line="310"/>
         <source>OPEN Root Access</source>
-        <translation type="unfinished"></translation>
+        <translation>Kök Erişimini AÇ</translation>
```
isaak654, line 1201: OPEN in this context means that the root access is open.

isaak654 (changes requested):

SandboxiePlus/SandMan/sandman_tr.ts
isaak654: "OPEN" in this context means that the root access is open.

isaak654 (approved):

Issue #2231, opened 2022-09-14
Sandboxie-Plus.ini keep old information


blap:
### Describe what you noticed and did

What I did:
Sandbox -> create new box
right click -> rename sandbox to whatever
or
right click -> remove sandbox

What I notice:
Sandboxie-Plus.ini keep the information of the old sandboxes names in "BoxDisplayOrder", "BoxCollapsedView" and "[SizeCache]". So I got random problems if I try to use the old names again.

### How often did you encounter it so far?

Always

### Affected program

Not relevant

### Download link

Not relevant

### Where is the program located?

The program is installed only inside a sandbox (NOT in the real system anyway).

### Expected behavior

Sandboxie-Plus.ini remove the old sandbox name that was renamed or deleted.

### What is your Windows edition and version?

Windows 10 Home 64 bits

### In which Windows account you have this problem?

Not relevant to my request.

### Please mention any installed security software

None

### What version of Sandboxie are you running?

Sandboxie-Plus-x64-v1.3.3.exe

### Is it a new installation of Sandboxie?

I recently did a new clean installation.

### In which sandbox type you have this problem?

In a Standard isolation sandbox (yellow sandbox icon).

### Can you reproduce this problem on an empty sandbox?

I can confirm it also on an empty sandbox.

### Did you previously enable some security policy settings outside Sandboxie?

None.

### Sandboxie.ini configuration

```shell
#
# Sandboxie-Plus configuration file
#

[GlobalSettings]
FileRootPath=C:\Users\Admin\0.Programas\Sandbox\%SANDBOX%
NetworkEnableWFP=y
KeyRootPath=\REGISTRY\USER\Sandbox_%USER%_%SANDBOX%
IpcRootPath=\Sandbox\%USER%\%SANDBOX%\Session_%SESSION%
Template=WindowsRasMan
Template=WindowsLive
Template=OfficeLicensing
Template=Edge_Win11Fix
Template=7zipShellEx
SandboxieLogon=y
InjectDll=...\logapi32.dll
InjectDll64=...\logapi64.dll
OpenPipePath=\Device\NamedPipe\LogAPI
StartRunAlertDenied=y
AlertProcess=EXCEL.EXE
AlertProcess=WINWORD.EXE
AlertProcess=POWERPNT.EXE
AlertProcess=notepad++.exe
AlertProcess=mpc-hc64.exe
AlertProcess=PDFXEdit.exe
AlertProcess=PDFXTools.exe
[...]

PR #2234, opened 2022-09-15
Cleanup configuration

okrc:
Fix https://github.com/sandboxie-plus/Sandboxie/issues/2231

### DIFF
--- a/SandboxiePlus/SandMan/SbiePlusAPI.cpp
+++ b/SandboxiePlus/SandMan/SbiePlusAPI.cpp
@@ -236,11 +236,8 @@ void CSandBoxPlus::UpdateSize()
 
 void CSandBoxPlus::SetSize(quint64 Size)
 { 
-	m_TotalSize = Size;
-	if (m_TotalSize == 0)
-		theConf->DelValue("SizeCache/" + m_Name);
-	else
-		theConf->SetValue("SizeCache/" + m_Name, Size);
+	m_TotalSize = Size; 
+	theConf->SetValue("SizeCache/" + m_Name, Size);
 }
 
 bool CSandBoxPlus::IsSizePending() const
--- a/SandboxiePlus/SandMan/SbiePlusAPI.cpp
+++ b/SandboxiePlus/SandMan/SbiePlusAPI.cpp
@@ -236,8 +236,11 @@ void CSandBoxPlus::UpdateSize()
 
 void CSandBoxPlus::SetSize(quint64 Size)
 { 
-	m_TotalSize = Size; 
-	theConf->SetValue("SizeCache/" + m_Name, Size);
+	m_TotalSize = Size;
+	if (m_TotalSize == 0)
+		theConf->DelValue("SizeCache/" + m_Name);
+	else
+		theConf->SetValue("SizeCache/" + m_Name, Size);
 }
 
 bool CSandBoxPlus::IsSizePending() const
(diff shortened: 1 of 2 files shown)


DavidXanatos:
[c19bda4](https://github.com/sandboxie-plus/Sandboxie/pull/2234/commits/c19bda441645beec7dec4adfc9ac7cd0b0ab9ad9)
and 
[3e406dc](https://github.com/sandboxie-plus/Sandboxie/pull/2234/commits/3e406dc4eebe5643a3febbba8f322184edff7c5d)
does not seam to be needed the box object being held by a smart pointer will be available until it goes out of scope so name does not need to be cached

Explicitly deleting SizeCache entries is an improvement, but it does not handle the case where a box is deleted manually from the ini or using the old classic ui. 
It would be best to add a clean up routine here https://github.com/sandboxie-plus/Sandboxie/blob/e5d1fab9edc40608fae853abf3f26c7ba47685b7/SandboxiePlus/SandMan/SandMan.cpp#L1600
such that on start/connect once we have a list of all existing boxes we purge the size cache.

okrc:
~~This PR doesn't do much more with SizeCache.~~
I will remove all changes to SizeCache.

> [c19bda4](https://github.com/sandboxie-plus/Sandboxie/pull/2234/commits/c19bda441645beec7dec4adfc9ac7cd0b0ab9ad9) and [3e406dc](https://github.com/sandboxie-plus/Sandboxie/pull/2234/commits/3e406dc4eebe5643a3febbba8f322184edff7c5d) does not seam to be needed the box object being held by a smart pointer will be available until it goes out of scope so name does not need to be cached
> 
> Explicitly deleting SizeCach

I don't understand here, is that [meant](https://github.com/sandboxie-plus/Sandboxie/pull/2234/commits/c19bda441645beec7dec4adfc9ac7cd0b0ab9ad9#diff-97100eacc2b039cb7dda7ed3ca92e6fb4572a58ab0a8db6c24bad0530c13a1dbL1236)?

https://github.com/sandboxie-plus/Sandboxie/pull/2234/commits/a8cdb4029593234d33d7b6d135e955369214a8cb
is an additional issue. The names are not determined and escaped for these matches.

okrc:
> I'll add this to the next build

这里还有一些问题，关闭统计后SizeCache依旧存在，但是Value已经不再更新。

There are still some problems here, after closing Count the SizeCache still exists, but the Value is no longer updated.

DavidXanatos:
I merged the changes and fixed a few things that did not seam right, also added an improved handling of the size cache

isaak654:
@DavidXanatos In PRs like this one, I would advise in the next future to use the "Squash and merge" type of merging instead of the traditional "Create a merge commit":
GitHub_merge

The first one is more chaotic when there is a certain number of commits in the same PR.

okrc:
@DavidXanatos
https://github.com/sandboxie-plus/Sandboxie/commit/52752774210f7fe168ae5a9175561430a1a82b08#
``` c++
			foreach(const QString & Key, theConf->ListKeys("SizeCache")) {
				if (!AllBoxes.contains(Key.toLower()))
					theConf->DelValue("SizeCache/" + Key);
			}
```

When the WatchBoxSize is modified, the handling should change here. For example:

``` c++
			if (theConf->GetBool("Options/WatchBoxSize", false)) {
				foreach(const QString & Key, theConf->ListKeys("SizeCache")) {
					if (!AllBoxes.contains(Key.toLower()))
						theConf->DelValue("SizeCache/" + Key);
				}
			}
			else {
				theConf->DelValue("SizeCache");
			}
```

DavidXanatos:
I'll go with 

```
			foreach(const QString & Key, theConf->ListKeys("SizeCache")) {
				if (!AllBoxes.contains(Key.toLower()) || !theConf->GetBool("Options/WatchBoxSize", false))
					theConf->DelValue("SizeCache/" + Key);
			}
```

PR #2241, opened 2022-09-17
Update sandman_zh_TW.ts

Hulen:
Update sandman_zh_TW.ts base sandman_en.ts


isaak654:
According to [Report-TradChinese.txt](https://github.com/sandboxie-plus/Sandboxie/blob/master/Sandboxie/msgs/report/Report-TradChinese.txt), [Text-TradChinese-1028.txt](https://github.com/sandboxie-plus/Sandboxie/blob/master/Sandboxie/msgs/Text-TradChinese-1028.txt) needs also an update.

PR #2243, opened 2022-09-18
Update Text-TradChinese-1028.txt

Hulen:
Update Text-TradChinese-1028.txt


isaak654:
Better to proceed a little at a time, otherwise the Parse script goes down.

Hulen:
OK, Thanks.

PR #2244, opened 2022-09-18
Update sandman_de.ts

bastik-1001:
Fresh fork to translate new strings and fix a grammar issue on an existing string.


bastik-1001:
I updated a second time, since the string is a denial of access, not an action sandboxie or the user does.

PR #2248, opened 2022-09-19
Clean up the invalid values that do not exist for m_Collapsed

okrc:
This should be the last change, it will clean up the invalid values that do not exist in m_Groups for m_Collapsed.


isaak654:
#2234

DavidXanatos:
not only groups can be collapsed but also boxes so the check must check for boxes and groups

okrc:
It occurred to me that the QStringList stored in m_Group contains the names of the group and the box. By eliminating the non-existent values from m_Collapsed in m_Group, the check is done for both the box and the group. Am I missing anything else?

okrc:
In fact, another omission here is used to clean up invalid Box names in m_Group.

PR #2249, opened 2022-09-19
Update Text-TradChinese-1028.txt

Hulen:
Update Text-TradChinese-1028.txt


Hulen:
I use WinMerge to compare English and Chinese Traditional archives for collation.

Maybe too many string.

isaak654:
The Korean translation file is slightly larger, so the Parse script is not able to process too many changes at the same time.
I would suggest to remove the commented out lines, but if you really want to restore them later, you could re-add them in different pull requests.

isaak654:
The Korean translation file is slightly larger, so the Parse script is not able to process too many changes at the same time.
I would suggest to remove the commented out lines introduced in this pull request, but if you really want to restore them later, you could re-add them in different pull requests.

Hulen:
OK

PR #2294, opened 2022-09-30
Update sandman_ko.ts

VenusGirl:
Update sandman_ko.ts 1.4.0

### DIFF
--- a/SandboxiePlus/SandMan/sandman_ko.ts
+++ b/SandboxiePlus/SandMan/sandman_ko.ts
@@ -4981,7 +4981,7 @@ To specify a process use &apos;$:program.exe&apos; as path.</source>
     <message>
         <location filename="Forms/OptionsWindow.ui" line="3226"/>
         <source></source>
-        <translation></translation>
+        <translation type="unfinished"></translation>
     </message>
     <message>
         <location filename="Forms/OptionsWindow.ui" line="3229"/>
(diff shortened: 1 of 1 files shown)


SandboxiePlus/SandMan/sandman_ko.ts
```diff
@@ -4767,230 +4767,231 @@ The process match level has a higher priority than the specificity and describes
...
         <location filename="Forms/OptionsWindow.ui" line="3197"/>
         <source>On File Recovery</source>
-        <translation type="unfinished"></translation>
+        <translation>파일 복구 시</translation>
     </message>
     <message>
         <location filename="Forms/OptionsWindow.ui" line="3226"/>
-        <source>This command will be run before a file is being recoverd and the file path will be passed as the first argument, if this command return something other than 0 the recovery will be blocked</source>
-        <translation type="unfinished"></translation>
+        <source></source>
```
isaak654, line 4983:
```suggestion
        <source>This command will be run before a file is being recoverd and the file path will be passed as the first argument, if this command return something other than 0 the recovery will be blocked</source>
```

Empty source is not a valid source.
Click on the 'Commit suggestion' button above if you agree with this change.

SandboxiePlus/SandMan/sandman_ko.ts
```diff
@@ -4767,230 +4767,231 @@ The process match level has a higher priority than the specificity and describes
...
+        <translation></translation>
```
isaak654, line 4984:
```suggestion
        <translation type="unfinished"></translation>
```

isaak654 (changes requested):

isaak654 (approved):

PR #2306, opened 2022-10-06
Fixed the display problem of Sandman.

okrc:
Problem triggered if the system color changes after Sandman starts in a dark color.

image

This issue has existed in multiple past versions.
I think these lines of code are redundant, but it brings up this display error. Please let me know if I've overlooked something.


DavidXanatos:
IIRC this lines of code are required to switch back from the "pizza" background

okrc:
> IIRC this lines of code are required to switch back from the "pizza" background

Is there a specific step? After commenting these lines of code, I tried to manually set the background to "pizza" and switch back, and I did not encounter an error.

DavidXanatos:
Indeed, i think there is a change to before, earlier it was posible to switch back from pizza without rebuilding the UI but now we just rebuild the UI for all major UI changes making that code obsolete

PR #2307, opened 2022-10-06
Changed the message notification for system color change.

okrc:


isaak654:
Out of curiosity - if David agrees, would you be available to become a collaborator in this repository?

okrc:
> Out of curiosity - if David agrees, would you be available to become a collaborator in this repository?

I'm sorry because my English is very poor and this may make some work very inefficient. I can only try to spend some time to make some improvements.

DavidXanatos:
does this message works for you? currently i don't hit that code no mater if the old or new message is monitored when i change the system preset for dark theme

okrc:
> currently i don't hit that code no mater if the old or new message is monitored when i change the system preset for dark theme

From the result of the translation, I don't understand the meaning of this sentence.

The old message actually only monitored the change in system theme color, i.e.
image
The new message is working on my Windows 11. It can handle dark and light color changes.

okrc:
Sorry, I tested Windows 10 and it didn't work. I should probably try WM_SETTINGCHANGE, which is used by some other projects.

DavidXanatos:
This seams to work with color 
how about we just monitor WM_SETTINGCHANGE that one seams to work just fine

PR #2308, opened 2022-10-06
Update sandman_de.ts

APMichael:
Just a few quick corrections.

### DIFF
--- a/SandboxiePlus/SandMan/sandman_de.ts
+++ b/SandboxiePlus/SandMan/sandman_de.ts
@@ -6313,7 +6313,7 @@ Bitte beachten Sie, dass diese Werte aktuell nutzerspezifisch sind und global f√
     <message>
         <location filename="Forms/SettingsWindow.ui" line="159"/>
         <source>Show the Recovery Window as Always on Top</source>
-        <translation>Zeige das Wiederherstellungsfenster immer im Vordergrund</translation>
+        <translation>Zeige das Wiederherstellungsfenster immer im Vordergrund an</translation>
     </message>
     <message>
         <location filename="Forms/SettingsWindow.ui" line="574"/>
@@ -6437,7 +6437,7 @@ Bitte beachten Sie, dass diese Werte aktuell nutzerspezifisch sind und global f√
     <message>
         <location filename="Forms/SettingsWindow.ui" line="196"/>
         <source>Show first recovery window when emptying sandboxes</source>
-        <translation>Zeige das Wiederherstellungsfenster vor dem Leeren der Sandboxen</translation>
+        <translation>Zeige das Wiederherstellungsfenster vor dem Leeren der Sandboxen an</translation>
     </message>
     <message>
         <source>Use Dark Theme (fully applied after a restart)</source>
--- a/SandboxiePlus/SandMan/sandman_de.ts
+++ b/SandboxiePlus/SandMan/sandman_de.ts
@@ -3564,7 +3564,7 @@ Notiz: Die Updateprüfung ist oft zeitversetzt zu den letzten GitHub-Veröffentl
     <message>
         <location filename="Windows/SettingsWindow.cpp" line="212"/>
         <source>Fully integrate</source>
-        <translation>Vollständig integriert</translation>
+        <translation>Vollständig integrieren</translation>
     </message>
     <message>
         <location filename="Windows/SettingsWindow.cpp" line="214"/>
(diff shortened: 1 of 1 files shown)


SandboxiePlus/SandMan/sandman_de.ts
```diff
@@ -3564,7 +3564,7 @@ Notiz: Die Updateprüfung ist oft zeitversetzt zu den letzten GitHub-Veröffentl
     <message>
         <location filename="Windows/SettingsWindow.cpp" line="212"/>
         <source>Fully integrate</source>
-        <translation>Vollständig integriert</translation>
+        <translation>Vollständig integrieren</translation>
```
bastik-1001, line 3567: That's probably correct, unless it would be 'fully integrated', which it isn't, but in either case it should work to be understood in the way it is supposed to be.

SandboxiePlus/SandMan/sandman_de.ts
```diff
@@ -5337,7 +5337,7 @@ Um eine Prozess anzugeben verwenden Sie &apos;$:program.exe&apos; als Pfad.</tra
     <message>
         <location filename="Forms/OptionsWindow.ui" line="2449"/>
         <source>Apply Close...=!&lt;program&gt;,... rules also to all binaries located in the sandbox.</source>
-        <translation>Wende Blockiert...=!&lt;program&gt;,... Regeln auch auf alle ausführbaren Binärcodes (Programme) innerhalb der Sandbox an.</translation>
+        <translation>Wende &apos;Close...=!&lt;program&gt;,...&apos; Regeln auch auf alle ausführbaren Binärcodes (Programme) innerhalb der Sandbox an.</translation>
```
bastik-1001, line 5340: Correct, since that's a command that does not seem to be supposed to be translated.

SandboxiePlus/SandMan/sandman_de.ts
```diff
@@ -6242,12 +6242,12 @@ Bitte beachten Sie, dass diese Werte aktuell nutzerspezifisch sind und global f
     <message>
         <location filename="Forms/SettingsWindow.ui" line="319"/>
         <source>Integrate boxes with Host Start Menu</source>
-        <translation>Integriere Boxen mit dem Startmenü des Hostrechners</translation>
+        <translation>Integriere Boxen in das Startmenü des Hostrechners</translation>
```
bastik-1001, line 6245: That reads indeed better.

SandboxiePlus/SandMan/sandman_de.ts
```diff
@@ -6437,7 +6437,7 @@ Bitte beachten Sie, dass diese Werte aktuell nutzerspezifisch sind und global f
     <message>
         <location filename="Forms/SettingsWindow.ui" line="196"/>
         <source>Show first recovery window when emptying sandboxes</source>
-        <translation>Zeige Wiederherstellungsfenster, vor dem Leeren der Sandboxen</translation>
+        <translation>Zeige das Wiederherstellungsfenster vor dem Leeren der Sandboxen</translation>
```
bastik-1001, line 6440: Shouldn't we add 'an', at the end, here as well? (I just tried to keep it short, but the change might be easier to read.)

Now I stumble upon the word 'first'. Can someone explain if that option does only apply if recovery is disabled, but the window can be enabled to be shown before deletion? I understood it phrased like "Show recovery window before emptying sandboxes" which would mean, to me, every time I empty a sandbox, I get to see a recovery window. The word first could mean I only ever get to see a recovery window for the first time, when I empty a box.

APMichael: You certainly understood that correctly and your translation should be correct as well.

The word "first" is meant here as "zuerst", as a sentence thus: "Zeige zuerst das Wiederherstellungsfenster an, wenn die Sandboxen geleert werden".

However, your translation is more understandable in my opinion.

SandboxiePlus/SandMan/sandman_de.ts
APMichael: The other selection option is: "Links nicht integrieren". Therefore I had corrected it.

isaak654:
When the PR is ready, please click the "Ready for review" button in order to switch the status.

APMichael:
> When the PR is ready, please click the "Ready for review" button in order to switch the status.

I think the PR can be merged. After all, it was only minor changes to single words.

PR #2348, opened 2022-10-15
Deprecating save-state and set-output commands

okrc:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


isaak654 (approved):
I definitely appreciate your efforts to update the workflows ([more than someone else](https://github.com/sandboxie-plus/Sandboxie/commit/c1d607fe7d9174f5dc2bb7313b066d3c4a9e9169)😥)

PR #1605, opened 2022-02-07
Create get_qttranslations.cmd to get translations files

DevSplash:
Issues: https://github.com/sandboxie-plus/Sandboxie/issues/1528
The translations files in QT 5.15.2 is incomplete, I tried to get them in QT 6.2.3. 
It worked for me, but I didn't (or couldn't) test the translations files for other languages.

Issue #1528, opened 2022-01-07
[Sbie Plus] Some Qt language files (zh_CN) are missing from the translations folder


isaak654:
**Describe the problem**
The following files are missing from `C:\Program Files\Sandboxie-Plus\translations`:
qtbase_zh_CN
qtscript_zh_CN
qtmultimedia_zh_CN
qtxmlpatterns_zh_CN

**To Reproduce**
1. Install Sandboxie Plus 1.0.7 x64
2. See the missing files above (I already tried a reinstall without success)
3. While the _zh_TW files are present, this seems only related to Simplified Chinese files.

**Expected behavior**
Those language files should match the ones listed [here](https://code.qt.io/cgit/qt/qttranslations.git/tree/translations?h=5.15-tr&id=10ac65e38651544104b32e5440c83271ea3f6175).

DavidXanatos:
these files seam to be missing from the official Qt distribution entirely :(

isaak654:
> these files seam to be missing from the official Qt distribution entirely :(

If it is really like that, then I'm speechless.

DevSplash:
In fact, I tested it last time by comparing and replacing files. Only the following three files are needed to solve the problem of incomplete translation of the current version.

![image](https://user-images.githubusercontent.com/15691402/148634268-4a7860d5-7e3a-4d76-8ecb-9a357db3a4f6.png)
- qt_zh_CN
- qtbase_zh_CN
- qtmultimedia_zh_CN

They can all be found in the translations directory of QT6.2.2. I didn't find any more missing translations after replacing the files.

isaak654:
In order to avoid future clutter, I would suggest to create a new .cmd to call [here](https://github.com/sandboxie-plus/Sandboxie/blob/master/Installer/merge_builds.cmd) to get those files directly from fixed URLs.

DevSplash:
This is probably because the translation file for QT 5.15.2 is incomplete. You can download the translation file for QT 6.2.2 from here.
https://download.qt.io/archive/qt/6.2/6.2.2/submodules/qttranslations-everywhere-src-6.2.2.zip
Need to send a cookie: `addr=ip address`

PR #2349, opened 2022-10-15
Remove get_qttranslations.cmd

isaak654:
It had already been renamed as fix_qt5_languages.cmd.

c.c. @DevSplash (original author of #1605)


isaak654:
#1528

Issue #2356, opened 2022-10-17
Firefox 106.0: error messages and crashes


APMichael:
### Describe what you noticed and did

The new Firefox version 106.0 does not work well with Sandboxie.

On my Windows 10 21H2 (x64) I get these error messages every time I watch a video:
```
PID 2216: SBIE2336 Error in GUI server: [55 / C0000155]
PID 2216: SBIE2335 Initialization failed for process WerFault.exe [88 / 0]
```
(PID 2216 = SbieSvc.exe)

On a Windows 11 21H2 (x64) in a VM, I get the following error message:
```
WerFault.exe (3064): SBIE2203 Failed to communicate with Sandboxie Service: *GUIPROXY_00000001; MsgId: 14 - WerFault.exe [FF000000]
```
In addition, in my tests in the VM, Firefox crashed very often on  e.g. YouTube or Bing.

The issue was already mentioned for the beta of Firefox 106.0:
https://www.wilderssecurity.com/threads/sandboxie-plus-v1-3-4.447639/page-2#post-3108293
https://www.wilderssecurity.com/threads/sandboxie-plus-v1-3-4.447639/page-3#post-3109121

And has also already been confirmed for the release version:
https://www.wilderssecurity.com/threads/sandboxie-plus-v1-4-0-v1-4-1-v1-4-2.447865/page-3#post-3110628

### How often did you encounter it so far?

Each time a video is viewed.

### Affected program

Firefox 106.0

### Download link

https://ftp.mozilla.org/pub/firefox/releases/106.0/win64/

### Where is the program located?

The program is installed only outside the sandbox.

### Expected behavior

As with previous Firefox versions: no error messages and crashes.

### What is your Windows edition and version?

Windows 10/11 Pro, 21H2, 64-bit

### In which Windows account you have this problem?

User account with UAC protection set to Always notify.

### Please mention any installed security software

Microsoft Defender Antivirus

### What version of Sandboxie are you running?

1.4.2

### Is it a new installation of Sandboxie?

I recently did a new clean installation.

### In which sandbox type you have this problem?

In a standard isolation sandbox (yellow sandbox icon).

### Can you reproduce this problem on an empty sandbox?

I can confirm it also on an empty sandbox.

APMichael:
By the way, I have seen this report #2288 as well.

Uninstalling and reinstalling Firefox did not bring any improvement. Also, the tests are always done in a clean VM, so Sandboxie and Firefox have to be freshly installed.

[...]

PR #2372, opened 2022-10-21
Fixes LibreWolf's addons not being included in the template

Dyras:
For whatever reason LibreWolf stores some of its addon data in the AppData Roaming folder, instead of storing it all in the Local AppData folder like Firefox does.

The end result? Even with access to "the entire profile folder" your addons are stored in the sandbox rather than outside of it.

This fixes that, so that LibreWolf users can once again enable "Auto delete content when last sandboxed process terminates" without having to reinstall all their addons afterwards.

### DIFF
--- a/Sandboxie/install/Templates.ini
+++ b/Sandboxie/install/Templates.ini
@@ -76,6 +76,7 @@ Tmpl.Waterfox=%AppData%\Waterfox\Profiles\*
 Tmpl.PaleMoon=%AppData%\Moonchild Productions\Pale Moon\Profiles\*
 Tmpl.SeaMonkey=%AppData%\Mozilla\SeaMonkey\Profiles\*
 Tmpl.LibreWolf=%AppData%\LibreWolf\Profiles\*
+Tmpl.LibreWolf_Extensions=%AppData%\LibreWolf\Extensions\*
 
 # Firefox-based addon path
 Tmpl.Zotero=%Tmpl.Firefox%\zotero
@@ -506,7 +507,7 @@ OpenFilePath=librewolf.exe,%Local AppData%\LibreWolf\Profiles\*\safebrowsing*
 Tmpl.Title=#4338,LibreWolf
 Tmpl.Class=WebBrowser
 OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\*
-OpenFilePath=librewolf.exe,%AppData%\librewolf*
+OpenFilePath=librewolf.exe,%Tmpl.LibreWolf_Extensions%\*
 
 #
 # Firefox Add-ons
--- a/Sandboxie/install/Templates.ini
+++ b/Sandboxie/install/Templates.ini
@@ -75,8 +75,7 @@ Tmpl.Firefox=%AppData%\Mozilla\Firefox\Profiles\*
 Tmpl.Waterfox=%AppData%\Waterfox\Profiles\*
 Tmpl.PaleMoon=%AppData%\Moonchild Productions\Pale Moon\Profiles\*
 Tmpl.SeaMonkey=%AppData%\Mozilla\SeaMonkey\Profiles\*
-Tmpl.LibreWolf=%AppData%\LibreWolf\Profiles\*
-Tmpl.LibreWolf_Extensions=%AppData%\LibreWolf\Extensions\*
+Tmpl.LibreWolf=%AppData%\LibreWolf\*
 
 # Firefox-based addon path
 Tmpl.Zotero=%Tmpl.Firefox%\zotero
@@ -477,37 +476,36 @@ ForceProcess=librewolf.exe
 [Template_LibreWolf_Bookmarks_DirectAccess]
 Tmpl.Title=#4336,LibreWolf
 Tmpl.Class=WebBrowser
-OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\bookmark*
-OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\places*
-OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\favicons.sqlite
+OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\Profiles\*\bookmark*
+OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\Profiles\*\places*
+OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\Profiles\*\favicons.sqlite
 
 [Template_LibreWolf_Cookies_DirectAccess]
 Tmpl.Title=#4328,LibreWolf
 Tmpl.Class=WebBrowser
-OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\cookies*
+OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\Profiles\*\cookies*
 
 [Template_LibreWolf_Passwords_DirectAccess]
 Tmpl.Title=#4331,LibreWolf
 Tmpl.Class=WebBrowser
-OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\logins.json
-OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\key*.db
+OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\Profiles\*\logins.json
+OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\Profiles\*\key*.db
 
 [Template_LibreWolf_Session_DirectAccess]
 Tmpl.Title=#4340,LibreWolf
 Tmpl.Class=WebBrowser
-OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\sessionstore.js*
+OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\Profiles\*\sessionstore.js*
 
 [Template_LibreWolf_Phishing_DirectAccess]
 Tmpl.Title=#4337,LibreWolf
 Tmpl.Class=WebBrowser
-OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\cert9.db
+OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\Profiles\*\cert9.db
 OpenFilePath=librewolf.exe,%Local AppData%\LibreWolf\Profiles\*\safebrowsing*
 
 [Template_LibreWolf_Profile_DirectAccess]
 Tmpl.Title=#4338,LibreWolf
 Tmpl.Class=WebBrowser
 OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\*
-OpenFilePath=librewolf.exe,%Tmpl.LibreWolf_Extensions%\*
 
 #
 # Firefox Add-ons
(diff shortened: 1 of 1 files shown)


isaak654:
https://github.com/sandboxie-plus/Sandboxie/blob/beb1ac8826759dfaeb18fcd78f78fde19a7e5561/Sandboxie/install/Templates.ini#L78
https://github.com/sandboxie-plus/Sandboxie/blob/beb1ac8826759dfaeb18fcd78f78fde19a7e5561/Sandboxie/install/Templates.ini#L508

I doubt that the problem is related to the Local AppData folder, because the current rule is equal to: `OpenFilePath=librewolf.exe,%AppData%\LibreWolf\Profiles\*`

isaak654:
https://github.com/sandboxie-plus/Sandboxie/blob/beb1ac8826759dfaeb18fcd78f78fde19a7e5561/Sandboxie/install/Templates.ini#L78
https://github.com/sandboxie-plus/Sandboxie/blob/beb1ac8826759dfaeb18fcd78f78fde19a7e5561/Sandboxie/install/Templates.ini#L508

I doubt that the problem is related to the Local AppData folder, because the current rule is equal to: `OpenFilePath=librewolf.exe,%AppData%\LibreWolf\Profiles\*\*`

> For whatever reason LibreWolf stores some of its addon data in the AppData Roaming folder

Please provide the exact path for the addon data, I would need a precise idea before approving the pull request.

Dyras:
The addons are stored here
`C:\Users\User\AppData\Roaming\librewolf\Extensions`
Perhaps the current rule could be edited to simply remove the Profiles part?
OpenFilePath=librewolf.exe,%AppData%\LibreWolf\*\*

isaak654:
In this case, I would be inclined to provide a `Tmpl.LibreWolf_Extensions=%AppData%\LibreWolf\Extensions\*` under the current `Tmpl.LibreWolf=%AppData%\LibreWolf\Profiles\*` and modify [this path](https://github.com/sandboxie-plus/Sandboxie/blob/beb1ac8826759dfaeb18fcd78f78fde19a7e5561/Sandboxie/install/Templates.ini#L509) accordingly.

This approach would allow the usage of [ExpandableVariables](https://sandboxie-plus.github.io/sandboxie-docs/Content/ExpandableVariables.html), especially the "Overriding Variables" part that could be useful in case of a sudden path change.

Please be sure to test the lines above for your use case, as I have no intention to install & testing LibreWolf.

Dyras:
We should be good to go now!

Unrelated: Do you happen to know if https://github.com/sandboxie-plus/Sandboxie/issues/2356 is a Firefox unique fix or if it affects all Firefox-based browsers? I reinstalled my computer because I got sick of having issues with Sandboxie-Plus and other other programs due to having a 5 year old Windows install, and now it would seem that LibreWolf is crashing with the exact message in that issue.

isaak654:
> Edit 4: I've figured it out. This line also needs to be added for the above to work `OpenFilePath=librewolf.exe,C:\Users\User\AppData\Roaming\librewolf\profiles.ini`
> 
> So what's your preferred method for this? Perhaps we should just remove the LibreWolf Extension and LibreWolf (Profile) templates and make a single template with this path? `C:\Users\User\AppData\Roaming\librewolf`

If I understood correctly, you would need:

1) to remove the lines below
https://github.com/sandboxie-plus/Sandboxie/blob/a969b663a7bdfff043a92b5f60250e2c63e65786/Sandboxie/install/Templates.ini#L79
https://github.com/sandboxie-plus/Sandboxie/blob/a969b663a7bdfff043a92b5f60250e2c63e65786/Sandboxie/install/Templates.ini#L510

2) to edit [this path](https://github.com/sandboxie-plus/Sandboxie/blob/a969b663a7bdfff043a92b5f60250e2c63e65786/Sandboxie/install/Templates.ini#L78) accordingly

Dyras:
The way I see it, I can do this in 2 ways:

### Way 1: Edit the template, remove LibreWolf_Extensions and it all works
Basically, I edit
`Tmpl.LibreWolf=%AppData%\LibreWolf\Profiles\*`

To
`Tmpl.LibreWolf=%AppData%\LibreWolf\*`

Advantages:
+ This will include all files for the LibreWolf profile
+ Probably more future proof, as it will include all future changes as well. Probably.

Disadvantages:
- Not as specific as the other ways

### Way 2: Edit all the templates so they fit

I replace
`Tmpl.LibreWolf=%AppData%\LibreWolf\Profiles\*`

To
`Tmpl.LibreWolf=%AppData%\LibreWolf\*`

Then I edit all the "lesser" templates to match. For example:
OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\bookmark*
Becomes
OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\Profiles\bookmark*


Advantages:
+ Will be more specific in which folders will be "Open" for Sandboxie

Disadvantages:
- More code changes
- Likely won't be as future proof

Tell me which way you'd prefer and I'll fix it right away

isaak654:
In my opinion:

- method 1 is preferable if the following issue could happen on other browsers as well:

> Edit 3: Hold on, something is still wonky about this. The addons aren't showing in a "Hardened Sandbox with Data Protection" box. It seems adding more files to the template will be necessary. I'll write later tonight when I figure out which ones.
>
> Edit 4: I've figured it out. This line also needs to be added for the above to work
`OpenFilePath=librewolf.exe,C:\Users\User\AppData\Roaming\librewolf\profiles.ini`

- method 2 is preferable if you are completely sure that this is only an issue exclusive to LibreWolf. Otherwise, massive changes would be needed for the other browsers too.

Dyras:
I've only noticed this in LibreWolf, so I assume it's specific to that browser. Will go with method 2 then.

isaak654:
Is the asterisk needed below?
https://github.com/sandboxie-plus/Sandboxie/blob/a242305d2eb487462375df61523311ade507b2a6/Sandboxie/install/Templates.ini#L505-L508

APMichael:
> Is the `\*` needed below?
> 
> https://github.com/sandboxie-plus/Sandboxie/blob/a242305d2eb487462375df61523311ade507b2a6/Sandboxie/install/Templates.ini#L505-L508

I would say yes, because the variable should be changed to:

`Tmpl.LibreWolf=%AppData%\LibreWolf`

Using `\*` in the variable is no longer useful with the shortened path, since it was only needed for the random profile ID.


Otherwise, when the variable is replaced, a wrong path is always created, e.g.:

`%AppData%\LibreWolf\*\Profiles\*\bookmark*`

Correct, however, would be:

`%AppData%\LibreWolf\Profiles\*\bookmark*`

isaak654:
> I can confirm, on the other hand, that the new full access (`%AppData%\LibreWolf\*`) is indeed required with a "Hardened Sandbox with Data Protection", but this then applies to Firefox (`%AppData%\Mozilla\Firefox\*`) and probably all other Firefox-based browsers. The reason why with a "Hardened Sandbox with Data Protection" the addons disappear is that the browser always creates a new profile if it cannot access the profiles.ini. Of course, the new profile will then not have access to the profile created outside the sandbox.

It seems that your objection confirms my previous concern in https://github.com/sandboxie-plus/Sandboxie/pull/2372#issuecomment-1288173857:
> - method 1 is preferable if the following issue could happen on other browsers as well:

So goodbye to method 2.

APMichael:
> It seems that your objection confirms my previous concern in [#2372 (comment)](https://github.com/sandboxie-plus/Sandboxie/pull/2372#issuecomment-1288173857):
> ...
> So goodbye to method 2.

I even think that the changes are not needed at all.

If someone uses a "Hardened Sandbox with Data Protection" and wants to use a profile outside the sandbox, then he must grant access to it via "NormalFilePath". This allows the templates to be used without changing them.

Examples:
`NormalFilePath=%AppData%\LibreWolf\*`
`NormalFilePath=%AppData%\Mozilla\Firefox\*`

In a "Standard Isolation Sandbox", everything works flawlessly with the existing template [Template_LibreWolf_Profile_DirectAccess] in my tests. So also for this case no changes should be necessary.

It would be great if @Dyras could double check this with his LibreWolf installation.

Dyras:
Addons do in fact show up as they should in a non-data protection sandbox for me. It would seem that the below makes it work as intended, with no other change to the original templates necessary:

```
[Template_LibreWolf_Profile_DirectAccess]
Tmpl.Title=#4338,LibreWolf
Tmpl.Class=WebBrowser
OpenFilePath=librewolf.exe,%Tmpl.LibreWolf%\*
NormalFilePath=librewolf.exe,%AppData%\LibreWolf\profiles.ini
```

Perhaps a similar approach should be added to every Firefox-based browser? I can make a different PR for it tomorrow if that would be better?

APMichael:
Thank you for the quick checking.

However, I must point out again that a "NormalFilePath" is only needed for a "Sandbox with Data Protection". All other sandboxes basically have read access (like "NormalFilePath=*") and an additional "NormalFilePath" is therefore never needed.

In my opinion, for this reason, a "NormalFilePath" should not be added to any of the existing browser templates. If a "Sandbox with Data Protection" is used, a "NormalFilePath" must be added manually to the sandbox settings.

Or new templates are added specifically for "Sandboxes with Data Protection", such as:

```
# Allow normal access to Firefox profile for sandboxes with data protection

[Template_Firefox_Profile_NormalAccess]
...
NormalFilePath=firefox.exe,%AppData%\Mozilla\Firefox\*
NormalFilePath=firefox.exe,%Local AppData%\Mozilla\Firefox\*
```
A user of a "Sandboxes with Data Protection" can then simply enable this new template and after that all already existing browser templates will also work as intended. (The question is whether this effort is worth it, because a "Sandboxes with Data Protection" will always require explanation, since almost everything outside the sandbox is blocked by default.)

Dyras:
Hmm well on the one hand, I agree that Data Protection is overkill for the average user and that you definitely need to look stuff up for that anyway. So it would make sense to add it as a separate template.

On the other hand, I think adding it to the default browser would be something we should consider for several reasons:
1. It's the most user friendly
2. It makes little to no difference for non-data protection users
3. The only ones that can use Data Protection are paid users, so I think having it as easy as possible will incentivize people to pay for the extra features.
4. Since Data Protection is for paid users only, adding an additional template for it will essentially be filler for most (non-paying) users.

Your thoughts @isaak654 ?

APMichael:
One last note from me on this topic (using Firefox as an example):

If the "NormalFilePaths" are to be added to the existing browser templates, then it should be kept in mind, that then all "Template_Firefox_xxx_DirectAccess" templates must have these 2 lines added to allow proper access to the browser profile outside of a "Sandbox with Data Protection":
```
NormalFilePath=firefox.exe,%AppData%\Mozilla\Firefox\*
NormalFilePath=firefox.exe,%Local AppData%\Mozilla\Firefox\*
```
Because you never know which and how many of the templates will be enabled by the user. To avoid such unnecessary multiple lines I had just suggested either to let the user add them manually or to add new separate templates.

But let's wait and see what @isaak654 or maybe @DavidXanatos think about it.

isaak654:
> To avoid such unnecessary multiple lines I had just suggested either to let the user add them manually or to add new separate templates.
>
> But let's wait and see what @isaak654 or maybe @DavidXanatos think about it.

I think writing new browser templates only to be applied for another sandbox type should be out of question.

If something has to be done, it would probably be more useful to introduce a sort of inline conditional statement:
`NormalFilePath=UsePrivacyMode|firefox.exe,%AppData%\Mozilla\Firefox\*` (only applied for sandboxes with data protection)

isaak654:
I think I wasn't clear enough, so I will try to explain better than I did.

We both share the same opinion about not changing the current Templates.ini behavior, I just wanted to throw a stone in order to offer a different perspective. 

Of course, a ton of `%AppData%\Mozilla\Firefox\*` \ `%Local AppData%\Mozilla\Firefox\*` paths for each Firefox template is an abuse that I would not recommend to anyone and the example I provided above is not something I wanted to be taken literally.

Personally I don't see anything wrong in settings working for a single box type, but you are free to think as you see fit.

PR #2499, opened 2022-11-26
Update Russian translation

lufog:
It took me some time to figure out why the program does not see the changes in the updated translation. Until I noticed the `translations.7z` archive in which I put the updated file.

As I understand it, now the `translations` folder is not needed, can it be deleted?


isaak654:
It wouldn't be the first time leftovers have been left out, hopefully more attention will be paid in the future.

I'm still on v1.5.3, so if it is as you say, I would like to suggest another pull request to fix this.

Find: https://github.com/sandboxie-plus/Sandboxie/blob/b126e8c6b1b92a8a7794a48c36e5e502a813e2f7/Installer/Sandboxie-Plus.iss#L81-L84

Replace all with: `Type: filesandordirs; Name: "{app}\translations"`

isaak654 (approved):

PR #2551, opened 2022-12-20
Add Contribution entries in Help menu

isaak654:
There is no explicit reference in the user interface about the contributor certificates, so let's put a contribution entry for both Plus and Classic UIs in the Help menu.

Plus UI help menu without a certificate applied:
Plus_help_no_cert

Plus UI help menu with a certificate applied:
Plus_help_cert

Classic UI help menu:
Classic_help


isaak654:
#1495

PR #2570, opened 2022-12-27
fixed issue with color inaccuracy caused by triggering OnColorSlider.

okrc:
When `color` is a definite color value, `m_BorderColor` is finally given an approximate value because `m_pColorSlider->setValue(hsv.h)` is triggered again.


okrc:
https://user-images.githubusercontent.com/29057533/209685426-4eae3e89-0628-4c07-adaa-acf8f299b201.mp4

PR #2634, opened 2023-01-26
Update sandman_de.ts

bastik-1001:
Another fresh fork to provide the translations for newly added strings. There are two commits as I went back to change the translation to have it reflect the uppercase start.

One string was not translated, as I was not sure, what `Copy empty` was supposed to mean. Since the original is as good as a faulty translation, I did not touch it, yet. My guess had been that it is an empty copy, which could be "Leerkopie", but that does not seem to fit the rest, where there is an option to always copy or never copy, so that I assume that it should be a verb, expressing what Sandboxie will do.

### DIFF
--- a/SandboxiePlus/SandMan/sandman_de.ts
+++ b/SandboxiePlus/SandMan/sandman_de.ts
@@ -642,12 +642,12 @@ Notiz: Die Updateprüfung ist oft zeitversetzt zu den letzten GitHub-Veröffentl
     <message>
         <location filename="Windows/OptionsGeneral.cpp" line="159"/>
         <source>Version 1</source>
-        <translation type="unfinished"></translation>
+        <translation>Version 1</translation>
     </message>
     <message>
         <location filename="Windows/OptionsGeneral.cpp" line="160"/>
         <source>Version 2</source>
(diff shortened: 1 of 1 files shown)


APMichael:
> One string was not translated, as I was not sure, what `Copy empty` was supposed to mean. ...

Ich denke es ist wirklich "Kopiere leer", im Sinne von "Kopiere die betreffende Datei, jedoch ohne ihren Dateiinhalt" gemeint. Eventuell kann man auch "Kopiere ohne Inhalt" oder "Kopiere ohne Daten" schreiben? Oder, damit es besser zu den beiden anderen Übersetzungen passt, vielleicht auch "Ohne Inhalt kopieren", "Ohne Daten kopieren" oder "Leer kopieren"?

Did you accidentally overlook the other new strings?

bastik-1001:
Danke, ich werde wohl "Ohne Inhalt kopieren" nehmen. Alles was du schreibst muss richtig sein, es wird eine Datei erzeugt, aber diese ist ohne Inhalt.

> Did you accidentally overlook the other new strings?

I did not touch any of them, as some contain byte values and I wasn't certain that those strings were finalized, but I will provide translations for them, shortly.

Issue #2685, opened 2023-02-12
"Add 'Run Sandboxed' to the explorer context menu" option does not work in ARM64 version


dnt1996:
### Describe what you noticed and did

1. Check the "Add 'Run Sandboxed' to the explorer context menu" option.
2. Click the OK button.
3. "explorer context menu" does not change.

### Affected program

Not relevant

### Download link

1.7.2

### Where is the program located?

The program is installed both inside and outside the sandbox.

### Expected behavior

-

### What is your Windows edition and version?

Windows 11 Pro 22H2

### In which Windows account you have this problem?

A local or Microsoft account without special changes.

### Please mention any installed security software

Defender

### What version of Sandboxie are you running?

1.7.2-arm64

### Is it a new installation of Sandboxie?

I recently did a new clean installation.

### In which sandbox type you have this problem?

In a standard isolation sandbox (yellow sandbox icon).

### Can you reproduce this problem on a new empty sandbox?

I can confirm it also on a new empty sandbox.

okrc:
Is there a `SbieShellPkg.msix` file in the installation directory?

#2415
https://github.com/sandboxie-plus/Sandboxie/pull/2359#issuecomment-1283973997

sepcnt:
1.7.2 also do not work on x64 (Windows 11)

dnt1996:
@okrc The file you mentioned does not exist.

okrc:
@dnt1996 https://github.com/sandboxie-plus/Sandboxie/issues/2447#issuecomment-1310329165

okrc:
@DavidXanatos Will the `SbieShellPkg.msix` file packaged by `x64\makeappx.exe` work on `ARM64` devices?
https://github.com/sandboxie-plus/Sandboxie/commit/9850c6d25864ac00ffe6a4b55c64c0b82a718596#diff-a81abec1c466dab0b09cb78f450059de78fe7fa04153b4243934d96307cedb92R324

DavidXanatos:
it should yes

PR #2690, opened 2023-02-12
Update SettingsWindow.cpp

sepcnt:
https://github.com/sandboxie-plus/Sandboxie/issues/2685


okrc (changes requested):
image

okrc (commented):
Removing `return` will cause the old context menu to be added to Windows 11 as well. they will appear at the same time, please modify the code.
移除 return，会使旧版上下文菜单也被添加到了  Windows 11。它们会同时出现，请修改代码。

PR #2749, opened 2023-03-14
Update Korean.isl

VenusGirl:


Installer/isl/Korean.isl
```diff
@@ -329,7 +329,7 @@ ErrorRegisterTypeLib=유형 라이브러리를 등록할 수 없습니다: %1
 
 ; *** Uninstall display name markings
 ; 예를 들어 '내 프로그램'으로 사용됩니다 (32비트)'
-UninstallDisplayNameMark=%1 (%2비트
+UninstallDisplayNameMark=%1 (%2)비트
```
isaak654, line 332: All other languages I checked have only `%1 (%2)` as value, why should it be different here?

VenusGirl (commented):
Error correction

Installer/isl/Korean.isl
VenusGirl: Format shown in source English.
")" was corrected because the parenthesis was missing.

; used for example as 'My Program (32-bit)'
UninstallDisplayNameMark=**%1 (%2)**
; used for example as 'My Program (32-bit, All users)'
UninstallDisplayNameMarks=**%1 (%2, %3)**

VenusGirl (commented):
")" was corrected because the parenthesis was missing.

PR #2764, opened 2023-03-25
Fix: Autorun long path's over 130 characters (including NUL character) not currently supported

Sapour:
This is a simple patch to add support for auto-run paths up to 32767 WCHAR's (including NUL character).

It also fixes the bug where _countof() was used instead of sizeof(), since RegQueryValueEx() expects the buffer size to be specified in bytes, not WCHAR's. Changing this on it's own would've increased max path length to 260 WCHAR's.

It's only for Sandboxie-Plus, but I plan on doing more comprehensive patch for everywhere MAX_PATH is currently used.

I have another patch which determine's required buffer size before-hand and only allocate's what's needed on the heap, but thought I'd keep it simple for starter's.

### DIFF
--- a/SandboxiePlus/SandMan/Helpers/WinAdmin.cpp
+++ b/SandboxiePlus/SandMan/Helpers/WinAdmin.cpp
@@ -110,19 +110,18 @@ bool IsAdminUser(bool OnlyFull)
 
 #define AUTO_RUN_KEY_NAME APP_NAME L"_AutoRun"
 
-constexpr size_t MAX_PATH_EX = 32767; // Long file path max length, in character's
+constexpr size_t MAX_PATH_EX = 32767; // Long file path max length, in characters
 
 bool IsAutorunEnabled()
 {
 	bool result = false;
 
 	HKEY hkey = nullptr;
-	if (RegOpenKeyEx (HKEY_CURRENT_USER, L"Software\\Microsoft\\Windows\\CurrentVersion\\Run", 0, KEY_ALL_ACCESS, &hkey) == ERROR_SUCCESS)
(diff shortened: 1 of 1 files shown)


DavidXanatos:
could you please update the code to allocate the 32k buffer on the heap instead of the stack

Sapour:
Can I ask why you prefer it on the heap ?

Also, if you prefer it on the heap, you might as well use my other patch, which determine's the exact size of buffer required before-hand. I'll get it up-loaded and submit the other patch, and you can decide which to merge.

SandboxiePlus/SandMan/Helpers/WinAdmin.cpp
```diff
@@ -110,15 +110,17 @@ bool IsAdminUser(bool OnlyFull)
 
 #define AUTO_RUN_KEY_NAME APP_NAME L"_AutoRun"
 
+constexpr size_t MAX_PATH_EX = 32767; // Long file path max length, in character's
```
isaak654, line 113:
```suggestion
constexpr size_t MAX_PATH_EX = 32767; // Long file path max length, in characters
```

isaak654 (changes requested):

DavidXanatos:
Given that stack space is limited I find it a good practice to nut put things which are multiple pages large in the stack.
In this case its rather fine as we know the expected call depth to reach this function.
But for other sandboxie functions this might be problematic, so its best not to start with it, code tends to diffuse around with time (through reusing etc...)

Sapour:
Thanks for info, I'll get both this PR updated, and also submit the other patch I mentioned which doesn't need a fixed size buffer.

You can then pick which one you think is better.

isaak654 (approved):

PR #2769, opened 2023-03-26
Fix: Autorun long path's over 130 characters (including NUL character) not supported.

Sapour:
Patch now supports unlimited characters (including NUL terminator), as the required buffer size is determined first, before allocating the buffer.

I also changed the check condition "ERROR_SUCCESS" to have that first, as it makes it clearer from first glance what you're checking against, instead of having to look to the end of the line. It also has the benefit when comparing against a variable, that it protect's against accidentally only using a single '=' sign instead of '==' - the former resulting in assignment, with result always being true if not zero. In this case it doesn't really apply, but I think it's preferable.

I'm happy to switch it back to the other way round if you want.

Also switched to using C++ style casting style, as it makes intent clearer and makes it easier to search for places where casting is taking place.

### DIFF
--- a/SandboxiePlus/SandMan/Helpers/WinAdmin.cpp
+++ b/SandboxiePlus/SandMan/Helpers/WinAdmin.cpp
@@ -115,14 +115,20 @@ bool IsAutorunEnabled()
 	bool result = false;
 
 	HKEY hkey = nullptr;
-	if (RegOpenKeyEx (HKEY_CURRENT_USER, L"Software\\Microsoft\\Windows\\CurrentVersion\\Run", 0, KEY_ALL_ACCESS, &hkey) == ERROR_SUCCESS)
+	if (ERROR_SUCCESS == RegOpenKeyEx (HKEY_CURRENT_USER, L"Software\\Microsoft\\Windows\\CurrentVersion\\Run", 0, KEY_ALL_ACCESS, &hkey))
 	{
-		WCHAR buffer[MAX_PATH] = {0};
-		DWORD size = _countof (buffer);
-
-		if (RegQueryValueEx (hkey, AUTO_RUN_KEY_NAME, nullptr, nullptr, (LPBYTE)buffer, &size) == ERROR_SUCCESS)
+		// First, determine the required buffer size, including NUL terminator (in bytes). RegGetValue() alway's adds
+		// an extra NULL terminator to size, even if one already exists, in case the stored value doesn't have one.
+		DWORD size {0};
+		if (ERROR_SUCCESS == RegGetValue(hkey, nullptr, AUTO_RUN_KEY_NAME, RRF_RT_REG_SZ, nullptr, nullptr, &size))
 		{
-			result = true; // todo: check path
+			// Then, allocate the buffer (in WCHARs) and retrieve the auto-run value. If successful, the size
+			// variable will be set to the actual size, without the extra NUL terminator.
+			auto buffer = std::make_unique< WCHAR[] >(size / sizeof(WCHAR));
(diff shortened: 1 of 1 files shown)


Sapour:
NOTE: You can either merge this one or the other one, which I'm about to update.

SandboxiePlus/SandMan/Helpers/WinAdmin.cpp
```diff
@@ -115,14 +115,20 @@ bool IsAutorunEnabled()
...
 
 	HKEY hkey = nullptr;
-	if (RegOpenKeyEx (HKEY_CURRENT_USER, L"Software\\Microsoft\\Windows\\CurrentVersion\\Run", 0, KEY_ALL_ACCESS, &hkey) == ERROR_SUCCESS)
+	if (ERROR_SUCCESS == RegOpenKeyEx (HKEY_CURRENT_USER, L"Software\\Microsoft\\Windows\\CurrentVersion\\Run", 0, KEY_ALL_ACCESS, &hkey))
 	{
-		WCHAR buffer[MAX_PATH] = {0};
-		DWORD size = _countof (buffer);
-
-		if (RegQueryValueEx (hkey, AUTO_RUN_KEY_NAME, nullptr, nullptr, (LPBYTE)buffer, &size) == ERROR_SUCCESS)
+		// First, determine the required buffer size, including NUL terminator (in bytes). RegGetValue() alway's adds
```
isaak654, line 120:
```suggestion
		// First, determine the required buffer size, including NULL terminator (in bytes). RegGetValue() always adds
```

isaak654 (changes requested):

SandboxiePlus/SandMan/Helpers/WinAdmin.cpp
```diff
@@ -115,14 +115,20 @@ bool IsAutorunEnabled()
...
+		// an extra NULL terminator to size, even if one already exists, in case the stored value doesn't have one.
+		DWORD size {0};
+		if (ERROR_SUCCESS == RegGetValue(hkey, nullptr, AUTO_RUN_KEY_NAME, RRF_RT_REG_SZ, nullptr, nullptr, &size))
 		{
-			result = true; // todo: check path
+			// Then, allocate the buffer (in WCHARs) and retrieve the auto-run value. If successful, the size
+			// variable will be set to the actual size, without the extra NUL terminator.
```
isaak654, line 126:
```suggestion
			// variable will be set to the actual size, without the extra NULL terminator.
```

isaak654 (changes requested):

SandboxiePlus/SandMan/Helpers/WinAdmin.cpp
Sapour: NUL is more consistent with definition (ie. a single character, which is what the terminator is).

[c++ null or nul](https://www.google.com/search?q=c%2B%2B+null+or+nul)

SandboxiePlus/SandMan/Helpers/WinAdmin.cpp
Sapour: Removed apostrophe (I do like them though :) ) but keeping at NUL as definition implies single character, which is what the terminator is.

Sapour:
Persoinally I think this patch is better than the other one as:

- it only uses required memory, but can handle any value size (up to available heap memory).
- Also, although 32767 is considered absolute limit, I've got a feeling it's not when environment variables are expanded in path, which this patch can also handle; the other #2764 patch can't.

However, it's highly unlikely to have paths this long though so, up to you.

PR #2771, opened 2023-03-27
Fix: Recovery: Only show delete confirmation dialog if file/folder selected

Sapour:
Fix: Recovery: Only show delete confirmation dialog if file/folder selected


DavidXanatos:
yea updating changelog to would be helpfull

Sapour:
Done.

PR #2808, opened 2023-04-10
Clean up of sandman_de.ts and typo fix for OptionsWindow.cpp comment

bastik-1001:
Fresh fork to provide a clean-up version of the German language file sandman_de.ts, by removing the entries that have vanished, plus some grammar improvement, since the upper-case spelling it is used that way ( [das] Erwzingen, where the first is and can be omitted.)

I also fixed a typo in the comment of a very recently committed change.

### DIFF
--- a/SandboxiePlus/SandMan/Windows/OptionsWindow.cpp
+++ b/SandboxiePlus/SandMan/Windows/OptionsWindow.cpp
@@ -135,7 +135,7 @@ public:
 	QSize sizeHint(const QStyleOptionViewItem &option, const QModelIndex &index) const
 	{
 		QSize size = QStyledItemDelegate::sizeHint(option, index);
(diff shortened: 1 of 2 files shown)


APMichael:
A cleanup is always good. 👍 What do you think about removing the <oldsource> lines as well? Most of the lines exist only because of minor typos.

bastik-1001:
> A cleanup is always good. +1 What do you think about removing the `<oldsource>` lines as well? Most of the lines exist only because of minor typos.

Aren't those from the source files that are used to generate the translations? In that case, they would be added back again, when the script updates the language files?

APMichael:
> Aren't those from the source files that are used to generate the translations? In that case, they would be added back again, when the script updates the language files?

I don't think so, since the number of `<oldsource>` lines varies widely in different translations.

bastik-1001:
Good, I created another pull request that removes those lines as well.

PR #2811, opened 2023-04-10
Update sandman_de.ts

bastik-1001:
Another fresh for to remove the old sources from sandman_de.ts


APMichael:
That was fast! 😃 Thank you very much.

PR #2812, opened 2023-04-10
Improve system check

isaak654:
- Replaced 6.4 version number (related to the obsolete Windows 10 Technical Previews) to 10.0
- Moved the x64 check to the beginning with an else in the middle
- Minor text variations

Note: if someone requires it, I don't have the chance to test the changes.


isaak654:
@sredna Any feedback or input would be greatly appreciated.

sredna:
Not personally a fan of string compares on versions but I'm sure it's fine.

If you are using the latest NSIS version you can use ```GetWinVer $outvarhere Build```

PR #2822, opened 2023-04-14
Update sandman_de.ts, typo fixes for recent commit

bastik-1001:
Fresh fork to provide new translations to the German strings in sandman_de.ts, plus two commits to fix typos made in comments.

It would be nice to have the strings reviewed, but I am open to make changes later on, as always.


bastik-1001:
I found and fixed mistakes I made in the translation file, leading to more commits. Unfortunately, I did not them while I was looking over them in Linguist or when I committed them in the first place.

PR #2825, opened 2023-04-15
Update sandman_de.ts

APMichael:
Minor fixes and changes.


bastik-1001:
Thank you for improving the readability on those translations, plus those fixes.

PR #2839, opened 2023-04-19
Update CHANGELOG.md

APMichael:
**Suggestion:** Make the changelog more readable by using nested lists.

**Edit:** I meant nested **lists**, of course, not nested tables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - added local template editor to the global settings window
 - added hidden message list
 - added option to prevent processes located outside the sandbox from loading boxed DLLs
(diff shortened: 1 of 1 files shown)


isaak654 (approved):

isaak654:
Personally I think this is brilliant, it remains only to convince @DavidXanatos.
If I merge this without his consent, we would definitely end up with future discrepancies.

APMichael:
@DavidXanatos is merging the pull request ok for you?

Unfortunately, GitHub's Markdown interpreter does not support the variant you are using:
```
- Item 1
-- Item 2
--- Item 3
```
But only this one:
```
- Item 1
  - Item 2
    - Item 3
```

APMichael:
Wow, that was fast. 🚀 Thanks a lot! 👍

PR #2848, opened 2023-04-23
Update sandman_de.ts and change-log spelling fix, plus minor changes

bastik-1001:
Fresh fork to translate the newly exposed strings. Spelling has been checked. The second commit adds a word to have it be correct, otherwise it would just have been the meaning of the word case. Unfortunately, that string is rather long when being translated.


APMichael:
You could possibly delete the 11 vanished translations, right?

bastik-1001:
Some of them might be outdated and could be removed, but those about the insider channel were new and I expect them to return, unless the idea would have been dropped. You fixed at least one of these entries. When similar strings appear again, I may use the string again, so there is no debate on how to word things.

The EFS entry might return as well. If not it will get removed at a later point, anyway.

Since I removed vanished entries not too long ago, I kept all of them. For some I am not even sure if they would need to be kept for reference sake.

APMichael:
Okay, that's fine with me.

I just mentioned it because the lines are often removed in other translations as well. Since you can always look at old commits in GitHub, nothing is ever lost and can be easily restored.

bastik-1001:
Just in case you need to some changes or translate something, I don't mind if you remove them.

You are right, there is always the commit history.

PR #2903, opened 2023-05-14
fix behaviour on multiple selection

okrc:
It should have been resolved now, and I haven't encountered any other issues locally for the time being. The two patches implement different behaviors. One of them prevents all selected items from moving further when the current position reaches a boundary, while the other terminates individually only when a boundary is reached within each group. I'm not sure which option to choose.

---
### Patch1 (This PR)
[0001-fix-behaviour-on-multiple-selection.patch](https://github.com/sandboxie-plus/Sandboxie/commit/e985c9beb8cdb70f1091757db00c2849a6c07202.patch)
### Patch2
[0002-fix-behaviour-on-multiple-selection.patch](https://github.com/sandboxie-plus/Sandboxie/files/11475270/0001-fix-behaviour-on-multiple-selection.patch)

### DIFF
--- a/SandboxiePlus/SandMan/Views/SbieView.cpp
+++ b/SandboxiePlus/SandMan/Views/SbieView.cpp
@@ -888,27 +888,39 @@ void CSbieView::OnGroupAction(QAction* Action)
 			return;
 
 		// todo: fix behaviour on multiple selection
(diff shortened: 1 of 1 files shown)


DavidXanatos:
the first simpler one sounds more intuitive

PR #2938, opened 2023-05-22
Additional Toolbar buttons

typpos:
Minor additions to Sandman toolbar: new box; new group; import box; empty all, refresh all.
Meta: Also an exercise for my workflow & setup. No objection to rejection.

### DIFF
--- a/SandboxiePlus/SandMan/SandMan.cpp
+++ b/SandboxiePlus/SandMan/SandMan.cpp
@@ -731,15 +731,14 @@ void CSandMan::CreateToolBar()
 	m_pToolBar = new QToolBar();
 	m_pMainLayout->insertWidget(0, m_pToolBar);
 
-	m_pToolBar->addAction(m_pMenuSettings);
+
+	m_pToolBar->addAction(m_pNewBox);
+	m_pToolBar->addAction(m_pNewGroup);
+	m_pToolBar->addAction(m_pImportBox);
 	m_pToolBar->addSeparator();
 
-	//m_pToolBar->addAction(m_pMenuNew);
-	//m_pToolBar->addAction(m_pMenuEmptyAll);
-	//m_pToolBar->addSeparator();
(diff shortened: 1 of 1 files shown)


DavidXanatos:
Can you marge the new box, new group, import into one button with  more options like its done for the clean up button.
and perhaps an UI to select which buttons show would be also great.

DavidXanatos:
Classically such configuration option is done using a windows with 2 lists where items from one can be added to the other for display and where teh order can also be changed.

I think though that this approach might be overkill for our few menu options, a right clock menu on the toolbar to enable individual buttons would surely be enough.

typpos:
I'll have a look at a configurable toolbar - it's a good learning exercise, but (subject to available time. I won't do a UI-driven full customization ( (a) limited time; (b) agreed overkill; (c) I thought this kind of obsessive customization feature died in the 2010s :-) ).

Plan, subject to change:
* Dropdown combo in global settings to select from a few templates ("None", "Default", "Compact", "All")  or type own sequence text, A B C | D E | F
* Entry in sanboxie plus ini: ToolbarItems=XXX

I'll do this in a separete PR and will close this one.

> Can you marge the new box, new group,

Lol. no ... and yes,  My main aim is to get to those functions with a single click and use all that massive dead UI real estate, not to create a duplicate squashed up dropdown menu that already exists (Menu > Sandbox). It's interesting how perspectives differ :-) . The above configurable toolbar will cover both preferences.

DavidXanatos:
A preset in settings is not optimal i think a context menu on the toolbar itself with a checkable for each possible tool bar entry would be better, also also quite easy to add.

thinking about further customizations we could allow users to pin RonCommand entries to the tool bar to start selected processes in preset boxes.
the ability to clear one particular box

typpos:
> a context menu on the toolbar itself ...

I like this idea.

I'll initially put an entry in the sandboxie-plus.ini without related UI-config, to manage my timeline. We can then go from there.

PR #2942, opened 2023-05-22
Update sandman_de.ts

bastik-1001:
Translated new strings.

Fresh fork.


bastik-1001:
No conflicts.

I wonder if the URLs should be placeholders instead of being in the string. That way, they don't need to be updated in every locale.

PR #2947, opened 2023-05-23
Fixed a command line issue when creating a process

lmou523:
Fixed command line issues caused by overmodifying the createprocessinternalw parameter. At some point, lpApplicationName needs to be null.

### DIFF
--- a/Sandboxie/core/dll/proc.c
+++ b/Sandboxie/core/dll/proc.c
@@ -857,7 +857,7 @@ _FX BOOL Proc_CreateProcessInternalW(
     BOOL ok;
     BOOL resume_thread = FALSE;
     WCHAR* lpAlteredCommandLine = NULL;
-    
+
     Proc_LastCreatedProcessHandle = NULL;
     LPCWSTR pSrcCommandLine = lpApplicationName;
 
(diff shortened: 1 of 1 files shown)


lmou523:
The first time I used pr,Please understand some irregularities.
The problem I found was that sandboxie changed lpApplicationName, causing excel exceptions.
In some cases, I restored the NULL value of lpApplicationName.
The changes are simple, and I made them directly on git.
Add "RecoverAppNameProcess=excel.exe" to the configuration file.

DavidXanatos:
can you describe a test case how to test that the new changes improve the behavior?

lmou523:
> can you describe a test case how to test that the new changes improve the behavior?

When installing office2016, open a word document，Run insert chart。
The #2947  is repeated(The Createprocessinternalw parameter lpApplicationName was changed when excel was started and should have been NULL)
Add "RecoverAppNameProcess=excel.exe" to the configuration file，The issue will be resolved，excel will start normally

DavidXanatos:
I have looked into the root issue and there is a more generic solution: https://github.com/sandboxie-plus/Sandboxie/commit/02d4e56b784c5e3e4d5f0290e6ff7f09515d45b5 which does not require user intervention, we just ensure the passed command-line is properly quoted.

lmou523:
> 我已经研究了根本问题，并且有一个更通用的解决方案：[02d4e56](https://github.com/sandboxie-plus/Sandboxie/commit/02d4e56b784c5e3e4d5f0290e6ff7f09515d45b5)，它不需要用户干预，我们只是确保正确引用传递的命令行。

lmou523:
> I have looked into the root issue and there is a more generic solution: [02d4e56](https://github.com/sandboxie-plus/Sandboxie/commit/02d4e56b784c5e3e4d5f0290e6ff7f09515d45b5) which does not require user intervention, we just ensure the passed command-line is properly quoted.

Thanks, I didn't verify the fix, I will test it when the next version of the sandbox is released

Issue #2890, opened 2023-05-10
Unable to open excel


lmou523:
### Describe what you noticed and did

Open Explorer, then double-click to open the excel file

The result is that excel is open, but the file is not

### How often did you encounter it so far?

every time

### Affected program

office 2016 32bit

### Download link

不相关

### Where is the program located?

The program is installed both inside and outside the sandbox.

### Expected behavior

The document should open normally

### What is your Windows edition and version?

windows 10 pro workstation 22h2

### In which Windows account you have this problem?

A local or Microsoft account without special changes.

### Please mention any installed security software

windows Defender

### What version of Sandboxie are you running?

sandboxie5.64.3

### Is it a new installation of Sandboxie?

I recently did a new clean installation.

### In which sandbox type you have this problem?

In a standard isolation sandbox (yellow sandbox icon).

### Can you reproduce this problem on a new empty sandbox?

I can confirm it also on a new empty sandbox.

### Sandboxie.ini configuration

```shell
#
# Sandboxie-Plus configuration file
#

[GlobalSettings]
Template=Edge_Win11Fix
Template=WindowsRasMan
Template=WindowsLive
Template=OfficeLicensing

[DefaultBox]
ConfigLevel=10
AutoRecover=y
BlockNetworkFiles=y
Template=OpenSmartCard
Template=OpenBluetooth
Template=SkipHook
Template=FileCopy
Template=qWave
Template=BlockPorts
Template=LingerPrograms
Template=AutoRecoverIgnore
RecoverFolder=%{374DE290-123F-4565-9164-39C4925E467B}%
RecoverFolder=%Personal%
RecoverFolder=%Desktop%
BorderColor=#00FFFF,ttl,6
Enabled=y

[UserSettings_176402EB]
SbieCtrl_UserName=teclink
SbieCtrl_BoxExpandedView=DefaultBox
SbieCtrl_NextUpdateCheck=1683719573
SbieCtrl_WindowCoords=1071,178,741,419
SbieCtrl_ActiveView=40021
SbieCtrl_ProcessViewColumnWidths=250,70,300
SbieCtrl_ReloadConfNotify=n
SbieCtrl_ExplorerNotify=n
SbieCtrl_EditConfNotify=n
SbieCtrl_HideWindowNotify=n
```

lmou523:
![image](https://github.com/sandboxie-plus/Sandboxie/assets/59944420/e3bcfc28-9900-48c9-800f-f74733159b79)
[...]

PR #2950, opened 2023-05-24
Handles double check of dde messages

lmou523:
In the default configuration, I opened Explorer, double-clicked an xlsx file, and excel opened, but the document did not.
The reason is that when excel processes dde messages, the handle to the service query is NULL,So I cancel the 32 bit double check.
office2016 64-bit office doesn't have this problem, but 32-bit office does（The operating system of my computer is x64）

#2890 
Recirculation step:
1.Install excel2016 32bit
2.Open explorer in the sandbox,Double-click an xlsx file
3.excel will open, but files will not


lmou523:
My ability is limited, I can only think of this way to solve the problem

PR #2961, opened 2023-05-25
Configurable toolbar items

typpos:
Related: #2938 

Feature:
* Ability to select what items are displayed in the advanced-mode toolbar of SB+.
* Right-click on toolbar for context menu to select items to be displayed
* Limited to actions that have an associated icon

Code review:
* Please, special attention to object allocation/destruction. Never ever worked with Qt before, and spoilt with years of C# garbage collection.

Thoughts:
* CSandMan actions should ideally be decoupled from Menu and be attached to main window. As before, toolbar kind of "borrows" them and assumes the menu is there and is complete.
* As before, the m_pSomeActionName members in CSandMan are a bit risky. Their validity is context based, not NULLed.

image

### DIFF
--- a/SandboxiePlus/SandMan/SandMan.cpp
+++ b/SandboxiePlus/SandMan/SandMan.cpp
@@ -360,7 +360,7 @@ void CSandMan::CreateUI()
 	m_pMainLayout->setSpacing(0);
 
 	if(iViewMode == 1)
-		CreateToolBar();
+		CreateToolBar(false);
 	else {
 		m_pSeparator = NULL;
 		CreateLabel();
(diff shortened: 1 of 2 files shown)


typpos:
Why is the ARM 64 build failing? Is this related to the change in Qt version? How do I correctly configure Qt on my local Win10x64 machine?

isaak654:
I think it's error C2593:

```
..\SandMan\SandMan.cpp(49): warning C4100: 'result': unreferenced formal parameter
..\SandMan\SandMan.cpp(439): warning C4100: 'bAdvanced': unreferenced formal parameter
..\SandMan\SandMan.cpp(822): error C2593: 'operator ==' is ambiguous
D:\a\Sandboxie\Qt\6.3.1\msvc2019_arm64\include\QtCore/qstring.h(1359): note: could be 'bool QString::operator ==(const char *) const'
D:\a\Sandboxie\Qt\6.3.1\msvc2019_arm64\include\QtCore/qstring.h(907): note: or       'bool operator ==(const QString &,std::nullptr_t) noexcept' [found using argument-dependent lookup]
D:\a\Sandboxie\Qt\6.3.1\msvc2019_arm64\include\QtCore/qstring.h(920): note: or       'bool operator ==(const QString &,const char16_t *) noexcept' [found using argument-dependent lookup]
..\SandMan\SandMan.cpp(822): note: while trying to match the argument list '(QString, int)'
..\SandMan\SandMan.cpp(874): error C2593: 'operator ==' is ambiguous
D:\a\Sandboxie\Qt\6.3.1\msvc2019_arm64\include\QtCore/qstring.h(1359): note: could be 'bool QString::operator ==(const char *) const'
D:\a\Sandboxie\Qt\6.3.1\msvc2019_arm64\include\QtCore/qstring.h(907): note: or       'bool operator ==(const QString &,std::nullptr_t) noexcept' [found using argument-dependent lookup]
D:\a\Sandboxie\Qt\6.3.1\msvc2019_arm64\include\QtCore/qstring.h(920): note: or       'bool operator ==(const QString &,const char16_t *) noexcept' [found using argument-dependent lookup]
..\SandMan\SandMan.cpp(874): note: while trying to match the argument list '(QString, int)'
..\SandMan\SandMan.cpp(2253): warning C4018: '<': signed/unsigned mismatch
```

typpos:
Thank you, @isaak654 !
Where do you find these logs? I couldn't find anything in the CI or local build.

DavidXanatos:
> Why is the ARM 64 build failing? Is this related to the change in Qt version? How do I correctly configure Qt on my local Win10x64 machine?


see here https://github.com/sandboxie-plus/Sandboxie/blob/master/SandboxiePlus/qmake_plus.cmd
you need to mod the target_qt.conf of the arm64 instalation

isaak654:
> Where do you find these logs? I couldn't find anything in the CI or local build.

My criteria was to look for `SandMan.cpp` in the failed job here:
https://github.com/sandboxie-plus/Sandboxie/actions/runs/5085381546/jobs/9138790312#step:15:895

The log in question will no longer be accessible after three months.

isaak654:
> Where do you find these logs? I couldn't find anything in the CI or local build.

My criteria was to look for `SandMan.cpp` in the failed job here:
https://github.com/sandboxie-plus/Sandboxie/actions/runs/5085381546/jobs/9138790312#step:15:895

The log in question will no longer be accessible for GitHub users after three months.

DavidXanatos:
Looks good, some comments though:
1. on a clean start without any presets the toolbar is completely empty, wo in such cases reset should be triggered autoamtically.
2. reset removes the notification label from the toolbar.
3. I think we don't need the setup wizard button 
4. I think we need buttons to "pause forcing programs", "disable file recovery" and "diable message popup"
5. I would utilize this menu buttons like used for cleanup more, may be put all the sandboxie.ini icons into one its really rare to edit the tempaltes or plus ini. Also perhaps merhe New boc new group and import into one such button
6. the first separator does not seam to be handled as well as the others when none of the 3 buttons is created it just stays there separating nothing.

I think this is really great progress and only needs a few small tweaks to be ready for the prime time.

typpos:
>  1. on a clean start without any presets the toolbar is completely empty, wo in such cases reset should be triggered autoamtically.

Fixed:
* If config contains no key "ToolBarItems", then default is used
* If config contains empty key "ToolBarItems", then toolbar is empty in line with user unchecking all toolbar items.

----
> 2. reset removes the notification label from the toolbar.

Not changed. The certificate is loaded from inconsistent locations when sandman.exe is not in the same folder as sbie and the certificate. 
* On start,  [main.cpp:26](https://github.com/typpos/Sandboxie/blob/7775d9106bbedadd4e4fedabe87763eb5e339118/SandboxiePlus/SandMan/main.cpp#L26) checks application directory, eg, debug/x64
* Label gets created without cert
* Real cert loaded by [SandMan.cpp:1991 ](https://github.com/typpos/Sandboxie/blob/7775d9106bbedadd4e4fedabe87763eb5e339118/SandboxiePlus/SandMan/SandMan.cpp#L1991) OnStatusChanged()
* Subsequent Toolbar re-builds have real cert present, so no label

When I config "UpdateAvailable=1.999", then the notification label remains visible when the toolbar is changed.

If this doesn't explain the issue you observed, then let me know, but I don't want to fiddle with the cert / updater / config-load logic.

----

> 3. I think we don't need the setup wizard button

Removed, though I think it should be left at user's discretion.

----

> 4. I think we need buttons to "pause forcing programs", "disable file recovery" and "diable message popup"

Only actions that have icons are shown. I added commented-out code. Once someone created icons, they can be uncommented.

----

> 5. I would utilize this menu buttons like used for cleanup moe, may be put all the sandboxie.ini icons into one its really rare to edit the tempaltes or plus ini. Also perhaps
   merhe New boc new group and import into one such button

Added menu buttons for New xxx and Edit xxx, but did not remove the individual on-toolbar-action buttons. The whole reason for me to do this feature is to have the actions flattened on the toolbar for single-click access, not in a 2-click menu. I hope that's an acceptable compromise.

----

> 6. the first separator does not seam to be handled as well as the others when none of the 3 buttons is created it just stays there separating nothing.

Bug. Fixed. Sorry.

----

Some suggested test cases:
* Config entries: None, empty, single, multiple, with invalid entry, with whitespaces
* button enabled-states: connect, disconnect, trace on/off, etc
* Toolbar config menu interaction: add, remove, reset, empty, all
* Views: (simple standard vintage) x (toolbar empty, single, multiple)
* Destruction of widgets with parents; no dangling or invalid use of pointers
* Removal of event connections: toggle, then remove trace, keep terminated, etc
* Toolbar label: cert, no cert, update available

PR #2968, opened 2023-05-26
Resolves #2963. Keep submenus visible on mouse leave.

### LINKED ISSUE #2963: Don't close submenus on mouse-leave

Win10x64
SB+ 1.9.3

A usability gripe with sandboxie is that submenus close rather quickly when the mouse does not hover. This makes it difficult to navigate deep nested menus in the application and in particular the traybar context menu.

Would it be ok to switch off the submenu timeout? I don't know if this is an accessibility feature or a cross-platform idea in Qt, but other applications on Windows don't exhibit this behavour.

Repro:
* tray icon context menu > Rclick > BoxName > Run > Standard Applications > (move mouse away from menu)
* menu closes

likewise for SB+ sub-menus.

On my system, the default timeout is set to 400ms. Maybe I'm slow or not precise enough...

Menus stay visible. There is no need to auto-close until they lose focus.

Something like so, but there might be a better solution...

SandMan\main.cpp

````
// Stop submenus from closing
class PreventSubMenuTimeoutStyle : public QProxyStyle {
public:
	PreventSubMenuTimeoutStyle(QStyle* style = 0) : QProxyStyle(style) {}

	int styleHint(StyleHint stylehint, const QStyleOption* opt = nullptr,
		const QWidget* widget = nullptr, QStyleHintReturn* returnData = nullptr) const
	{
		if (stylehint == SH_Menu_SubMenuSloppyCloseTimeout) return -1;
		return QProxyStyle::styleHint(stylehint, opt, widget, returnData);
	}
};

...
QtSingleApplication app(argc, argv);
app.setQuitOnLastWindowClosed(false);

// qDebug() << "default timeout equals " << app.style()->styleHint(QStyle::SH_Menu_SubMenuSloppyCloseTimeout);
auto st = new PreventSubMenuTimeoutStyle(app.style());
app.setStyle(st);
````

### PR DESCRIPTION
**Pull request by typpos**
Feature:
* Don't hide sub-menus on mouse leave.

Review:
* May require testing on platforms other than Win10x64.

### DIFF
--- a/SandboxiePlus/SandMan/Helpers/KeepSubMenusVisibleStyle.h
+++ b/SandboxiePlus/SandMan/Helpers/KeepSubMenusVisibleStyle.h
@@ -1,0 +1,16 @@
+#pragma once
+
+#include <QProxyStyle>
+
+// Keep submenus open when mouse leaves
+class KeepSubMenusVisibleStyle : public QProxyStyle {
+public:
+	KeepSubMenusVisibleStyle(QStyle* style = 0) : QProxyStyle(style) {}
+
+	int styleHint(StyleHint styleHint, const QStyleOption* opt = nullptr,
+		const QWidget* widget = nullptr, QStyleHintReturn* returnData = nullptr) const
+	{
+		if (styleHint == SH_Menu_SubMenuDontStartSloppyOnLeave) return 1;
+		return QProxyStyle::styleHint(styleHint, opt, widget, returnData);
+	}
+};
--- a/SandboxiePlus/SandMan/SandMan.cpp
+++ b/SandboxiePlus/SandMan/SandMan.cpp
@@ -35,6 +35,7 @@
 #include <QSessionManager>
 #include "Helpers/FullScreen.h"
 #include "Helpers/WinHelper.h"
+#include "Helpers/KeepSubMenusVisibleStyle.h"
 
 CSbiePlusAPI* theAPI = NULL;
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by DavidXanatos** [comment]
the fix as is works good under Qt6, but when built with Qt5 it breaks the side tabs:
grafik

with the proxy style set globally on the app we never reach

```
void CustomTabStyle::drawControl(ControlElement element, const QStyleOption* option, QPainter* painter, const QWidget* widget) const {
	if (element == CE_TabBarTabLabel ...
```
such that the if would be satisfied

Its not apparent on a first glance why this fails, also given that with Qt6 it works just fine... perhaps its a Qt5 bug.

I would suggest to rework the code to apply the style on the menus created directly, instead of on the entire app.

typpos:
Investigated and I think this should be closed unless someone has a simple and clean means to do this.

Notes for psterity:

I can't seem to set styles on menus to affect the timeout:
* QMenu constructor doesn't let style be set and picks it from app
* QMenu::QMenu() calls QMenuPrivate::init which calls QMenuSloppyState::initialize(), which caches m_dont_start_time_on_leave and other values from style in member variables. These do not appear to get updated when QMenu style is changed later. So I can change, for example, the background color of a menu, but not the timeout.

A workaround is to set the app style just before menus are built, and then reset it to the previous value. 
* Refer to my above commit as an example.
* This doesn't look clean and needs to be done for every menu creation
* It may have side-effects on other platforms or where stylesheets are used
* Requires setParent() juggling for custom style instance so that there is no crash when it's removed from app while menu still exists. Example commit keeps it alive & dangling under CSandMan instance.

Tried but can't get to the SloppyMenu timer id to kill timer in event loop.

Admittedly my total Qt experience is a few hours so I might still learn a thing or two.

isaak654:
> Its not apparent on a first glance why this fails, also given that with Qt6 it works just fine... perhaps its a Qt5 bug.

Qt 5 is not going to last forever, so looking on the bright side [this commit](https://github.com/sandboxie-plus/Sandboxie/pull/2968/commits/44ae34ca9388e1d4bae5c45c93da147deee6a882) may come in handy sooner or later.

DavidXanatos:
I found a way to fix that on qt5 now the question do we dear to put it in 1.10.4 final or do we delay it for 1.11.0 preview?

PR #3030, opened 2023-06-09
Ability to select font for ini editor panels.

typpos:
For consideration...

Feature:
* Ability to change font used in the ini-editor panels for box-options and general-settings
* Ability to reset the font to default

Notes:
* New SB+ ini entry:  UIConfig/IniFont
  * Default if missing or empty or blank
  * Otherwise the string format as per QFont.toString()
* Code duplication for Options & Settings Windows. Could be made a separate handler/provider class.
* No broadcast if multiple box-options windows are open

Fine if rejected.

image


DavidXanatos:
Do you think you could add this to the context menu of the text edit instead of adding buttons?
or may be putt it here: 
grafik

Somehow the setting buttons in that place look somewhat out of place, some users will surely appreciate the ability to change fonts, but where the buttons are right now is a bit misplaced.

typpos:
Moved font selection to "Display Options" tab. 

Font change applies immediately, not when clikcing "Apply".

image

PR #3038, opened 2023-06-14
FixDirLinkErr

lmou523:
Fixed a link error when the path was created with a symbolic link, which could cause a software exception


DavidXanatos:
could you please provide a test case how to trigger the bug

PR #3044, opened 2023-06-16
Improve deletion of the Sandboxie folder

isaak654:
The uninstaller of Sandboxie Classic tends to leave the Sandboxie installation folder as empty in some cases, so I would suggest to further improve the deletion.

### DIFF
--- a/Sandboxie/install/SandboxieVS.nsi
+++ b/Sandboxie/install/SandboxieVS.nsi
@@ -849,10 +849,10 @@ Function CheckUpdates
 DoCheck:
   DetailPrint "Running UpdUtil ..."
   SetDetailsPrint listonly
-  
+
   ExecWait '"$INSTDIR\UpdUtil.exe" $0 sandboxie /step:scan /scope:meta /version:${VERSION}' $1
 	;MessageBox MB_OK "UpdUtil: $0"
-	
+
   IntCmp $1 0 is0 lessthan0 morethan0
   is0:
     ;DetailPrint "no update"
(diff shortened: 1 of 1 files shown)


isaak654:
@sredna Any feedback or input would be greatly appreciated.

sredna:
RMDir (without /r) only deletes a directory if it is empty, you don't need a function. If it failed to delete then maybe something has a open handle, maybe call RMDir and Sleep in a loop while IfErrors is true after RMDir. Give up after 5 seconds.

Process Monitor will tell you why the delete fails.

DeleteDirIfEmpty is also broken, the returned order of . and .. is undefined.

isaak654:
> Process Monitor will tell you why the delete fails.

I've just tried, but I'm unable to reproduce this issue consistently. However, I would still like to provide a minimal fix that doesn't break anything.

> If it failed to delete then maybe something has a open handle, maybe call RMDir and Sleep in a loop while IfErrors is true after RMDir. Give up after 5 seconds.

Please check my refreshed commit. 

Also, I would like to know if setting a `SetOutPath $TEMP` before the first `RMDir "$INSTDIR"` could further help as in [this example](https://github.com/rime/weasel/blob/8ca8142e12f72eda191c9ab80e4c25ff7168c8d2/output/install.nsi#L280).

sredna:
> Please check my refreshed commit

You need a clearerrors before the first of the two rmdir's.

> Also, I would like to know if setting a SetOutPath $TEMP before the first RMDir "$INSTDIR" could further help

It will not hurt anything. It is used because it calls SetCurrentDirectory and the current directory holds a handle. If the current directory was the problem then it should happen every time.

Your code is a bit strange with everything in functions and no uninstaller section?

isaak654:
Thanks for your report, a new commit was pushed.

> Your code is a bit strange with everything in functions and no uninstaller section?

Frankly I'm not sure what was the intention of the original developer, basically the code now belongs to the Sandboxie community: https://news.sophos.com/en-us/2020/04/09/sandboxie-is-now-an-open-source-tool/

Sorry if I'm rambling a little bit, but I think what we really miss at this time is compatibility support for UWP apps and at least an independent security audit provided by a security research team... I'm grateful with the over 20 security issues fixed internally since the open-source release, but it's a bit unfair that other open-source software like [Veracrypt](https://sourceforge.net/p/veracrypt/discussion/general/thread/1a04e8bf82/) receive spontaneous attention in this regard.

PR #3071, opened 2023-07-03
Add CodeQL workflow

jorgectf:
Hello from [GitHub Security Lab](https://securitylab.github.com/)!

Your repository is critical to the security of the Open Source Software (OSS) ecosystem, and as part of our mission to make OSS safer, we are contributing a [CodeQL configuration for code scanning](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning-for-a-repository#setting-up-code-scanning-manually) to your repository. By enabling code scanning with CodeQL, you will be able to continuously analyze your code and surface potential vulnerabilities [before they can even reach your codebase](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/triaging-code-scanning-alerts-in-pull-requests#about-code-scanning-results-on-pull-requests).

We’ve tested the configuration manually before opening this pull request and adjusted it to the needs of your particular repository, but feel free to tweak it further! Check [this page](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#editing-a-code-scanning-workflow) for detailed documentation.

Questions? Check out the FAQ below!

### FAQ
<details>
<summary>Click here to expand the FAQ section</summary>

#### How often will the code scanning analysis run?
By default, code scanning will trigger a scan with the CodeQL engine on the following events:
* On every pull request — to flag up potential security problems for you to investigate before merging a PR.
* On every push to your default branch and other protected branches — this keeps the analysis results on your repository’s *Security* tab up to date.
* Once a week at a fixed time — to make sure you benefit from the latest updated security analysis even when no code was committed or PRs were opened.

#### What will this cost?
Nothing! The CodeQL engine will run inside GitHub Actions, making use of your [unlimited free compute minutes for public repositories](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#about-billing-for-github-actions).

#### Where can I see the results of the analysis?
The results of the analysis will be available on the *Security* tab of your repository. You can find more information about the results [here](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/managing-code-scanning-alerts-for-your-repository#viewing-the-alerts-for-a-repository).

For Pull Requests, you can find the results of the analysis in the *Checks* tab. You can find more information about the Pull Request results [here](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/triaging-code-scanning-alerts-in-pull-requests#about-code-scanning-as-a-pull-request-check).

#### What types of problems does CodeQL find?
CodeQL queries are hosted in the [`github/codeql`](https://github.com/github/codeql) repository.

By default, code scanning runs the [`default` query suite](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/built-in-codeql-query-suites#default-query-suite). The queries in the `default` query suite are highly precise and return few false positive code scanning results.

If you are looking for a more comprehensive analysis, which could return a greater number of false positives, you can enable the [`security-extended` query suite](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/built-in-codeql-query-suites#security-extended-query-suite) in the `queries` option of `github/codeql-action/init`.

In the event of finding a false positive, please create a [false positive Issue](https://github.com/github/codeql/issues/new?assignees=&labels=false-positive&projects=&template=ql--false-positive.md&title=False+positive) in [`github/codeql`](https://github.com/github/codeql) so we can investigate and improve the query in question. You can also contribute to the query by opening a pull request against [`github/codeql`](https://github.com/github/codeql).

#### How do I customize the analysis?
You can customize the analysis by using a CodeQL configuration file. This way, you can specify which queries should [not] be run, and/or which files should be excluded from the analysis. You can find more information about the configuration file [here](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/customizing-code-scanning#using-a-custom-configuration-file).

#### How do I upgrade my CodeQL engine?
No need! New versions of the CodeQL analysis are constantly deployed on GitHub.com; your repository will automatically benefit from the most recently released version.

#### The analysis doesn’t seem to be working
If you get an error in GitHub Actions that indicates that CodeQL wasn’t able to analyze your code, please [follow the instructions here](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/troubleshooting-the-codeql-workflow) to debug the analysis.

#### Which source code hosting platforms does code scanning support?
GitHub code scanning is deeply integrated within GitHub itself. If you’d like to scan source code that is hosted elsewhere, we suggest that you create a mirror of that code on GitHub.

</details>

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,3 @@
+query-filters:
+  - exclude:
+      id: cpp/incorrect-string-type-conversion
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -58,6 +58,7 @@ jobs:
       uses: github/codeql-action/init@v2
       with:
         languages: ${{ matrix.language }}
+        config-file: ./.github/codeql/codeql-config.yml
         # If you wish to specify custom queries, you can do so here or in a config file.
         # By default, queries listed here will override any specified in a config file.
         # Prefix the list here with "+" to use these queries and those in the config file.
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,83 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ 'master' ]
+  pull_request:
+    # The branches below must be a subset of the branches above.
+    branches: [ 'master' ]
+  schedule:
+    - cron: '33 7 * * 6'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: windows-2019
+    timeout-minutes: 45
+
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'cpp' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
+        # Use only 'java' to analyze code written in Java, Kotlin or both.
+        # Use only 'javascript' to analyze code written in JavaScript, TypeScript or both.
+        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Setup msbuild
+      uses: microsoft/setup-msbuild@v1
+
+    # Qt 6 is only required for ARM64 build currently.
+    # - name: Install Qt6 x64
+    #   uses: jurplel/install-qt-action@v3
+    #   with:
+    #     version: '6.3.1'
+    #     arch:  'win64_msvc2019_64'
+
+    - name: Install Qt5 x64
+      uses: jurplel/install-qt-action@v3
+      with:
+        version: '5.15.2'
+        arch:  'win64_msvc2019_64'
+        cache: true
+
+    - name: Installing Jom
+      run: SandboxiePlus\install_jom.cmd
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v2
+      with:
+        languages: ${{ matrix.language }}
+        # If you wish to specify custom queries, you can do so here or in a config file.
+        # By default, queries listed here will override any specified in a config file.
+        # Prefix the list here with "+" to use these queries and those in the config file.
+
+        # For more details on CodeQL's query packs, refer to: https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs
+        # queries: security-extended,security-and-quality
+
+    - name: Build Sandboxie x64
+      run: msbuild /t:build Sandboxie\Sandbox.sln /p:Configuration="SbieRelease" /p:Platform=x64 -maxcpucount:8
+
+    - name: Build Sandboxie-Plus x64
+      run: SandboxiePlus\qmake_plus.cmd x64
+
+    - name: Build SbieShell x64
+      run: msbuild /t:restore,build -p:RestorePackagesConfig=true SandboxiePlus\SbieShell\SbieShell.sln /p:Configuration="Release" /p:Platform=x64
+
+    - name: Build Sandboxie-Tools x64
+      run: msbuild /t:build SandboxieTools\SandboxieTools.sln /p:Configuration="Release" /p:Platform=x64 -maxcpucount:8
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v2
+      with:
+        category: "/language:${{matrix.language}}"


DavidXanatos:
So this is intended for code analysis, why does the workflow then need to build all the projects? given that our CI run takes about 30 minutes this seams wasteful, also Qt6 is only required for ARM64 build currently so if this only needs to build the x64 version that dependency could be omitted.

jorgectf:
Hi @DavidXanatos, thanks for your comment! I am not 100% familiar with Sandboxie, so please feel free to remove any step that you think should not be there 👍. I suggest you let the workflow build what you want CodeQL to analyze.

isaak654:
> also Qt6 is only required for ARM64 build currently so if this only needs to build the x64 version that dependency could be omitted.

I've just commented out that part.

@jorgectf Do you know if there is a specific support channel for asking to reduce the build time for [main.yml](https://github.com/sandboxie-plus/Sandboxie/blob/master/.github/workflows/main.yml)? As you can notice there, we even had to set `timeout-minutes: 45`, so this is a huge concern for us.

jorgectf:
> @jorgectf Do you know if there is a specific support channel for asking to reduce the build time for [main.yml](https://github.com/sandboxie-plus/Sandboxie/blob/master/.github/workflows/main.yml)? As you can notice there, we even had to set `timeout-minutes: 45`, so this is a huge concern for us.

You can also set the `timeout-minutes: 45` directive in this workflow, as it's a GitHub Actions workflow as well.

isaak654:
> > @jorgectf Do you know if there is a specific support channel for asking to reduce the build time for [main.yml](https://github.com/sandboxie-plus/Sandboxie/blob/master/.github/workflows/main.yml)? As you can notice there, we even had to set `timeout-minutes: 45`, so this is a huge concern for us.
> 
> You can also set the `timeout-minutes: 45` directive in this workflow, as it's a GitHub Actions workflow as well.

Of course, no problem with it.

We would, however, be interested in reducing the build time. In case you know where to direct us to improve the situation, we remain at disposal.

DavidXanatos:
I see a lot of "Cast from char* to wchar_t*" complaints, these casts are on purpose as we usually define any generic buffer as BYTE* i.e. unsigned char* as we want to use byte wise pointer arithmetic on them, how to make the scanner understand that and not complain? Do we have to now always cast to void* before casting that to wchar_t* ?

jorgectf:
> We would, however, be interested in reducing the build time. In case you know where to direct us to improve the situation, we remain at disposal.

Ah! Well, that would fall into GitHub Actions itself. I suggest you explore https://support.github.com/features/actions to see where to direct your feedback. If you can't find the right place to write, let me know!

> I see a lot of "Cast from char* to wchar_t*" complaints, these casts are on purpose as we usually define any generic buffer as BYTE* i.e. unsigned char* as we want to use byte wise pointer arithmetic on them, how to make the scanner understand that and not complain? Do we have to now always cast to void* before casting that to wchar_t* ?

I have disabled the query in question in https://github.com/sandboxie-plus/Sandboxie/pull/3071/commits/779de999953f2ab857161708a37b7ec7b0d10cf8, however, I suggest you expand the explanation for the query, as it might surface a security/quality problem.

DavidXanatos:
ok thank you lets see what this run reports, I think when the casting issue is fixed we can use it :D

isaak654:
[main.yml](https://github.com/sandboxie-plus/Sandboxie/actions/runs/5443914158/jobs/9901079154?pr=3071) just failed without any change on our side, so I think something else should be improved.

May I suggest to exclude some unnecessary files from the CodeQL scan (on both pull and push events)?

```
    paths-ignore:
      - '**/*.md'
      - '**/*.txt'
      - '**/*.yml'
```

jorgectf:
Sure! You are free to change anything. However, take into account that, for compiled languages, [`paths-ignore` does not really work](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/customizing-code-scanning#specifying-directories-to-scan), it depends on what the workflow builds between `/init` and `/analyze`. It's very probable that those files are not being scanned at all.

DavidXanatos:
@jorgectf 
how to fix that: "Setting a DACL to NULL in a SECURITY_DESCRIPTOR will result in an unprotected object." we want unprotected object on purpose, can we disable this warning like the wchar_t* cats one?

isaak654:
I would personally opt to dismiss the alerts when there are only a couple of false positives that cannot be fixed manually in the code.

Any exclusions you add in codeql-config.yml should only be reserved for a large number of reports that cannot be fixed manually in the code... in this case, you should be able to find the corresponding ID once you access any query in this list: https://codeql.github.com/codeql-query-help/cpp/

DavidXanatos:
yea sounds more reasonable, i have dismissed that two warnings and added an explanatory comment, now all alerts are fixed

@jorgectf thanks for the contribution

PR #3082, opened 2023-07-06
A few typo fixes

nkh0472:


isaak654 (changes requested):
From https://github.com/sandboxie-plus/Sandboxie/discussions/1123

> ⚠️ You shouldn't ever edit the text inside the source tags of the .ts file(s) before applying the same change on the respective source code file(s). For example, in commit [3b86441](https://github.com/sandboxie-plus/Sandboxie/commit/3b864411275ac0fa6e60ea1544feb176b470c049) I edited the source file and all the .ts files together. In your case, you could just update your specific .ts file together with the source file(s) in question and nothing else.

nkh0472:
>>  ⚠️ You shouldn't ever edit the text inside the source tags of the .ts file(s) before applying the same change on the respective source code file(s). For example, in commit [3b86441](https://github.com/sandboxie-plus/Sandboxie/commit/3b864411275ac0fa6e60ea1544feb176b470c049) I edited the source file and all the .ts files together. In your case, you could just update your specific .ts file together with the source file(s) in question and nothing else.

Thanks for noticing :)

isaak654 (approved):

PR #3092, opened 2023-07-11
More build instructions

derivativeoflog7:
More build instructions to build Sandboxie, and once I figure it out for Sandboxie-Plus too

I can't get to build for ARM64 on my machine though
image


isaak654:
As far as I know, there is no ARM64 build for Sandboxie Classic.

DerivativeOfLog7:
So uhm... I restarted VS2019, without touching anything or changing any component... and now it just works???
image

isaak654:
There is no public ARM64 build available for Sandboxie Classic, so any update to improve the build instructions on this front is welcome.

I also believe the huge number of warnings in Visual Studio should be faced somehow... is there any chance they could slow down the build time?

DerivativeOfLog7:
> There is no public ARM64 build available for Sandboxie Classic, so any update to improve the build instructions on this front is welcome.

Some core components (like the driver) are contained in Sandbox.sln, so at least part of that has to be compiled anyway. I'd love to test the GUI of Sandboxie classic on ARM64, but I don't have a device with that... I'll check with emulators hoping it won't be slow as molasses

> I also believe the huge number of warnings in Visual Studio should be faced somehow... is there any chance they could slow down the build time?

Likely not, warnings are mostly about code that could cause issues in some edge cases

DerivativeOfLog7:
I can't seem to find a way to target ARM64 on VS, is it not possible for now?

DavidXanatos:
which VS version are you using?
it shoudl be possible, if oyu are missing that option you need to start the setup and add the required components arm64 and arm64ec which may be still marked as experimental

DerivativeOfLog7:
> which VS version are you using? it shoudl be possible, if oyu are missing that option you need to start the setup and add the required components arm64 and arm64ec which may be still marked as experimental

Forgot to say, I can't seem to target ARM64 on SandboxiePlus.sln
image

I can target ([and compile for](https://github.com/sandboxie-plus/Sandboxie/pull/3092#issuecomment-1630934015)) ARM64 on Sandbox.sln
image

DavidXanatos:
I started with a RaspberyPi 4 and then moved to a ARM64 laptop IdeaPad 5g,
you can add additional targets using the configuration manager of the solution.

DerivativeOfLog7:
@isaak654 fair point, I think it even can be moved to the top as qmake_plus.bat doesn't use curl or 7z

Also, sorry for being absent these days, I need to study for an upcoming exam so I can't dedicate much time to other things

DerivativeOfLog7:
@DavidXanatos wdyt about adding an argument to  qmake_plus.cmd to make it possible to change the Qt directory? It could even be optional defaulting to the current value as not to break current workflows, but would make it possible for people who have not the same directory structure to compile without having to resort to making symlinks to the Qt directory

DerivativeOfLog7:
So there seems to be an issue: I need to use the <_arch_> Native Tool Command Prompt for VS2019 to compile SBP for Win32/x64 (where arch is the respective target architecture), since the "normal" developer command prompt doesn't work... but that means that I can't test building for ARM64 on my machine
Let me know if this is normal

DerivativeOfLog7:
well of course the check fails... the QT version is mismatched

isaak654:
@DerivativeOfLog7 Considering the delay in evaluating your update request to Qt 6.3.2, could I create a new pull request containing only the changes in [Sandboxie/ReadMe.md](https://github.com/sandboxie-plus/Sandboxie/pull/3092/files#diff-d7c26afa5bf327943f4380640ece6821a0dbd22a7c1ccfca84ce6e9a29125329) and [SandboxiePlus/SandMan/SandMan.vcxproj](https://github.com/sandboxie-plus/Sandboxie/pull/3092/files#diff-08540406573b52b460182614744ce14c4f2cc801c76b87e4304d0f5d1c0bb78d)?

Kizuno18:
This branch is now superseded by the current build scripts and cannot be merged safely as a small conflict resolution.

Current `master` centralizes Qt versions in `Installer/buildVariables.cmd` (`6.8.3`), uses the MSVC 2022 layouts, selects the Qt 6 project explicitly, and contains later rewrites of `install_jom.cmd`. The relevant later commits include `a307347c`, `17aa8306`, and `9b80690d`.

There is also a concrete argument-contract conflict: current CI invokes `SandboxiePlus\\qmake_plus.cmd x64 build_qt6` and `ARM64 build_qt6`, while this branch treats argument 2 as `qt_root_dir`. Rebasing this implementation would therefore use `build_qt6` as a filesystem root and break both active CI calls. The branch still pins Qt 5.15.13/6.3.2 and `msvc2019*` paths as well.

I recommend closing this draft as superseded. Any remaining request for a configurable Qt root or jom parallelism should be implemented separately against the current argument contract and toolchain.

PR #3210, opened 2023-08-16
Update sandman_de.ts

APMichael:
- Added new translations.
- Removed "vanished" messages.
---
- **Comment:** I think translating the different certificate types is not necessary, otherwise it will only lead to confusion.
- **Help:** Any idea for the translation of "**node** locked"?

### DIFF
--- a/SandboxiePlus/SandMan/sandman_de.ts
+++ b/SandboxiePlus/SandMan/sandman_de.ts
@@ -2849,7 +2849,7 @@ Nein wählt: %2</translation>
     <message>
         <location filename="SandMan.cpp" line="2600"/>
         <source>The Certificate is node locked.</source>
-        <translation>Das Zertifikat ist gesperrt.</translation>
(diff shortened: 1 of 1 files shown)


isaak654:
> **Help:** Any idea for the translation of "**node** locked"?

In the changelog, there is a reference to hardware bound certificates.

APMichael:
> In the changelog, there is a reference to hardware bound certificates.

Thank you for the hint. You are right, it refers to the hardware-bound certificates.

https://en.wikipedia.org/wiki/Node-locked_licensing

bastik-1001:
> * **Comment:** I think translating the different certificate types is not necessary, otherwise it will only lead to confusion.

Translating different certificate types/names, can indeed lead to confusion. For that to work better, they would need to be placeholders. Every other string mentioning them needs to pull the name from the placeholder. I am not sure if there is a need to have them localized in the first place. 

There are strings that can't be translated, like what seems to be the structure of the product key. Depending on how strict the parsing is, it will lead to issues. Having left to right languages could be a problem, if they use to structure it the other way around.

PR #3265, opened 2023-08-30
Changelog 1.11.1

NewKidOnTheBlock:


bastik-1001:
In line 29, ImDisk Driver is spelled with an uppercase D for the word driver, while in line 27 the word driver is spelled with a lowercase d.

> added ImDisk driver integration (...)

should probably be the correct wording, since the driver needs to be installed for Sandboxie to interact with it. So not the driver itself was added, but the integration of support to utilize the driver. As far as I am aware.

NewKidOnTheBlock:
Good catch. I'm on it.

PR #3283, opened 2023-09-07
Update Templates.ini

offhub:
Allows the AdGuard Browser Assistant extension to communicate with the AdGuard desktop program.


DavidXanatos:
is this always teh same string:
OpenIpcPath=Adguard.BrowserExtensionHost.exe,*\BaseNamedObjects*\net.pipe:?bmV0LnBpcGU6Ly8rL0FER1VBUkRBUElFTkRQT0lOVC8=

offhub:
According to [this page](https://learn.microsoft.com/en-gb/archive/blogs/rodneyviana/named-pipes-in-wcf-are-named-but-not-by-you-and-how-to-find-the-actual-windows-object-name) it should be the same.

PR #3390, opened 2023-10-21
Update german translation

bastik-1001:
Fresh fork to translate the new strings and to update the newly added troubleshooting translation for the German language.

I choose "Datenträger" for "Volume", as this should be more easily understood, compared to "Volumen", which is technically valid, but has other meanings as well.

### DIFF
--- a/SandboxiePlus/SandMan/Troubleshooting/lang_de.json
+++ b/SandboxiePlus/SandMan/Troubleshooting/lang_de.json
@@ -1,10 +1,10 @@
 {
-"Description Text...": "Beschreibungs Text...",
-"Fix current issues": "Aktuelle problemem beheben",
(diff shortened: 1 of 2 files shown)


APMichael:
**lang_de.json**

"Wählen Sie, welche Box in eine Box mit reduziertem Isolations-App-Fach umgewandelt werden soll."

Sounds strange, I would suggest the following:

"Wählen Sie, welche Box in eine Applikationsunterteilungsbox mit reduzierter Isolation umgewandelt werden soll."

bastik-1001:
You are right, I applied your suggested wording.

The original was David's version, and I was not sure, how to word it, much better. Thank you for your help.

DavidXanatos:
I used chat GPT for the translation template LOL

PR #3413, opened 2023-10-29
Update sandman_de.ts

bastik-1001:
Added the new strings and removed the old vanished entry.

The second commit fixes a typo and while seeing the removed entry, I shortened the translation by omitting the word web, which is not required.

### DIFF
--- a/SandboxiePlus/SandMan/sandman_de.ts
+++ b/SandboxiePlus/SandMan/sandman_de.ts
@@ -3265,7 +3265,7 @@ Fehler: %1</translation>
     <message>
         <location filename="SandMan.cpp" line="3936"/>
         <source>Do you want to open %1 in a sandboxed or unsandboxed Web browser?</source>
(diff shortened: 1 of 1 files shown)


SandboxiePlus/SandMan/sandman_de.ts
```diff
@@ -3265,17 +3265,17 @@ Fehler: %1</translation>
     <message>
         <location filename="SandMan.cpp" line="3936"/>
         <source>Do you want to open %1 in a sandboxed or unsandboxed Web browser?</source>
-        <translation type="unfinished"></translation>
+        <translation>Möchten Sie %1 in einem sandgeboxten or nicht sandgeboxten Browser öffnen?</translation>
```
APMichael, line 3268: You accidentally overlooked the "or".

bastik-1001: Thanks.

(That makes it longer again by two characters. :) )

PR #3436, opened 2023-11-16
[Chromium] Update browser paths

isaak654:
1) Updated most Chromium templates ending with "Cookies_DirectAccess" , here the directory `\Network` must be added before `\Cookies*` (except the entry related to [Osiris browser](https://github.com/sandboxie-plus/Sandboxie/pull/3436#issuecomment-1817672861))

2) The `\Default` directory must be appended to the "Tmpl.Opera" variable

3) Fixed some trailing spaces

4) Updated browser template wizard for Chromium

### DIFF
--- a/Sandboxie/install/Templates.ini
+++ b/Sandboxie/install/Templates.ini
@@ -87,7 +87,7 @@ Tmpl.Chrome=%Local AppData%\Google\Chrome\User Data\Default
 Tmpl.Edge=%Local AppData%\Microsoft\Edge\User Data\Default
 Tmpl.Vivaldi=%Local AppData%\Vivaldi\User Data\Default
 Tmpl.Brave=%Local AppData%\BraveSoftware\Brave-Browser\User Data\Default
-Tmpl.Opera=%AppData%\Opera Software\Opera Stable
+Tmpl.Opera=%AppData%\Opera Software\Opera Stable\Default
 Tmpl.Yandex=%Local AppData%\Yandex\YandexBrowser\User Data\Default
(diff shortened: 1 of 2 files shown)


APMichael:
I assume that this needs to be adapted for all Chromium-based browsers. (I can confirm this for Microsoft Edge). And as far as I know, Opera now also uses the "Default" directory, so the variable "Tmpl.Opera" would need to be updated.

isaak654:
> I assume that this needs to be adapted for all Chromium-based browsers. (I can confirm this for Microsoft Edge).

The only exception is the Osiris browser maintained by [Acent](https://twitter.com/Acent_tech/status/1457677696868052994) which has not been updated for several months... considering their installer doesn't allow to download any update, it should be excluded for the time being.

DavidXanatos:
so how did the paths changed exactly? we will need to add it to the browser template wizard as well

APMichael:
@isaak654

Thank you very much for your research.

----

@DavidXanatos

It only affects "Cookies_DirectAccess", here the directory "\Network" must be added before the directory "\Cookies*". 

And the "\Default" directory must be appended to the "Tmpl.Opera" variable.

isaak654:
> so how did the paths changed exactly? we will need to add it to the browser template wizard as well

@DavidXanatos 
I've added a modified TemplateWizard.cpp in this pull request, feel free to check.

isaak654:
> so how did the paths changed exactly? we will need to add it to the browser template wizard as well

@DavidXanatos 
I've added a modified TemplateWizard.cpp in this pull request, feel free to check it out.

PR #3449, opened 2023-11-27
Update lang_de.json

APMichael:
- Added new translations.
- Minor changes to translations (to achieve better consistency).


APMichael:
Thanks for merging.

For whatever reason, "BOM" was lost due to my update. Is there an easy way to change the encoding from "UTF-8" back to "UTF-8-BOM"?

PR #3493, opened 2023-12-14
Move Plus license to the repository root

isaak654:
- [x] `LICENSE` file has been renamed to `LICENSE.Classic`.
- [x] Plus file `license.txt` has been renamed to `LICENSE.Plus` and moved to the repository root. 
- [x] A symbolic link that points to `LICENSE.Plus` has been placed into the `Installer` folder. 
Note: this works when you browse the source code in the GitHub Web UI:
immagine.

With these changes, GitHub is able to detect both Classic and Plus licenses at the same time:

licenses_2


isaak654:
#3495

PR #3595, opened 2024-02-03
Discord_Server

kokofixcomputers:
Add join discord server


kokofixcomputers:
Changes the invite to infinity

isaak654 (approved):

DavidXanatos:
i added a non expiering link i think it looks better when the invite says David Invites you

kokofixcomputers:
> i added a non expiering link i think it looks better when the invite says David Invites you

No problem! Thank you!

PR #3603, opened 2024-02-05
Changelog fixes

NewKidOnTheBlock:


NewKidOnTheBlock:
Orange boxes were sometimes referred to as "Security Enhanced" but they are called "Security Hardened" in the wizard.
Also, there was a mention of "a black box". This could be confusing to people who don't know that it refers to encrypted boxes.

PR #3607, opened 2024-02-06
Update SECURITY.md and add discord support tickets

kokofixcomputers:
Update SECURITY.md and add discord support tickets

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,7 +2,7 @@
 
 ## Reporting a vulnerability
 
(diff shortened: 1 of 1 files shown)


kokofixcomputers:
> I'm not sure if we should use discord for security reports i think githubs mechanism is good and reliable, and security researchers will have a github account already anyways probably.

oh... well there is a support ticket thing

kokofixcomputers:
okay then

Issue #3584, opened 2024-01-28
Add "On Terminate" trigger


e-t-l:
### Is your feature request related to a problem or use case?

Just as the trigger "On Box Start" very helpfully runs commands when the first program starts in a box, an "On Terminate" trigger would run commands when programs in a box **terminate**.

### Describe the solution you'd like

It would suffice to add a general "On Box Terminate" trigger that runs when all programs have terminated, but if possible a more granular "On _Program_ Terminate" that runs on the termination of specific programs would be great!

Sandboxie has a Stop Behavior feature, so there's already infrastructure for performing actions based on program termination events. This proposal would simply provide an interface for users to add their own actions.

### Describe alternatives you've considered

AutoHotKey is the only other solution I can think of.

bastik-1001:
I think that a trigger that runs something once the last process in a box has terminated, is very useful.

On the aspect of granularity, I am not sure how difficult that is, for some software, like those that spawn child processes, maybe even with different names or when two instances of the same software are run, but it appears indeed similar to the stop behavior, so that the same logic applies.

PR #3642, opened 2024-02-25
This submission completes the foreground UI and the data writing part…

love-code-yeyixiao:
[issues3584](https://github.com/sandboxie-plus/Sandboxie/issues/3584)
This submission completes the foreground UI and the data writing part of the“On Terminate” trigger, but it remains to complete the actual trigger part.


love-code-yeyixiao:
Then what am I supposed to do? Where are triggers formally parsed and processed?

Issue #1985, opened 2022-06-24
Prevent sandboxed programs from accessing screenshots of a window outside the sandbox


czoins:
### Is your feature request related to a problem or use case?

Any process running inside the sandbox can take screenshots/record your screen, increasing risk when you cannot disable internet access for the process (for usability)

### Describe the solution you'd like

Is it possible to prevent processes running inside the box from having access to the screen/other windows? Other products like Kaspersky implement this feature, and return a black screen instead of the actual content when taking a screenshot/screen recording. Just like the clipboard access, I don't think (from my limited knowledge) that the processes need this access to function.

isaak654:
Duplicate of #624

DavidXanatos:
Please elaborate which otehr products except kaspersky,a nd how do thay do it in which usecases etc.

DavidXanatos:
This feature has been added in 1.13.4 however please consider that it is user mode protection only hence malware targeting sandboxie explicitly can bypass it, also the implementation may not yet cover all means available to capture the screen.

Issue #624, opened 2021-02-28
Option to block screen capture inside the sandbox


0x4E69676874466F78:
See: https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowdisplayaffinity
Example code:
https://github.com/Cingozr/SetWindowDisplayAffinity
https://github.com/ysc3839/NoScreenCap
At the moment, any application in the sandbox can get a fullscreen shot and thus get the contents of the windows of other sandboxes (and not just sandboxes).

DavidXanatos:
these examples only show how to protect one window globaly its used now to protect encrypted black boxe content, but it does nto help to iplement the oposit, i.e. whats requested here.
Does anyone have some workign examples how to globaly block screeshoting?

0x4E69676874466F78:
Yes it's just an example of system functionality for specific windows. 
This is only if we want to protect the windows of a particular sandbox. 

I haven't dug into the API, but in your case you probably need to hijack the function to call the screenshot of the screen.

bastik-1001:
`BlockScreenCapture=y` and `CoverBoxedWindows=y` have been added months ago, the former aims at preventing sandboxed apps capturing the screen and the latter tries to hide sandboxed apps from being captured. So, even if there might be cases in which it can't be used, I consider this to be implemented.

PR #3670, opened 2024-03-03
Prevents processes in the sandbox from capturing screen images.

love-code-yeyixiao:
By hooking on Apis such as GetDC, attempts to get an out-sandbox window HDC or a desktop window HDC are intercepted.
https://github.com/sandboxie-plus/Sandboxie/issues/624
https://github.com/sandboxie-plus/Sandboxie/issues/1985

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - added menu entry to restart Sandman as admin [#3581](https://github.com/sandboxie-plus/Sandboxie/issues/3581) (thx Yeyixiao)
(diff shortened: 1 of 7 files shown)


offhub:
`IsBlockCapture=y`

Many programs do not work properly when this setting is enabled. (Stuck, crash, missing ui elements)

- explorer.exe
- Chromium based things
- Start.exe

Start_lUtBk82fQt

DavidXanatos:
@love-code-yeyixiao this seams to be broken indeed i have rolledback the changes in the current version,
this needs to be fixed before it can go public, also afaik on can also screenshot using directx api which here is not handles.
could you please revise the implementation acordingly

PR #3692, opened 2024-03-10
Update sandman_de.ts

bastik-1001:
Just translated the new string. I kept the vanished entry, assuming that it will be required again, once the code is ready to perform the protection.


bastik-1001:
My initial goal to separate the language file update and the changes to spelling mistakes, by creating a branch, did fail, as each modification was a new branch, even though I used the same name, but GitHub made new ones, appending numbers. Then I failed again to merge them to a single branch that was not the main branch. This results in now having only one pull request, instead of the intended two.

PR #3693, opened 2024-03-10
Add“Prevent the program from getting the image of  un-sandboxed  windows” function

love-code-yeyixiao:
UI was added now.

Fixed a simplified Chinese translation.


love-code-yeyixiao:
It has been tested and is working fine, but has yet to be perfected. In addition, the DirectX Way also needs to consider, but this matter leaves others to do, I am tired now.

love-code-yeyixiao:
[#1985 ](https://github.com/sandboxie-plus/Sandboxie/issues/1985)

PR #3707, opened 2024-03-16
Update README.md

bastik-1001:
I thought it to be nice to mention some more changes, that occurred since Sandboxie had its sources released.

The contribution of Yeyixiao, should probably be included in the section, but I don't know what to write.


isaak654:
> The contribution of Yeyixiao, should probably be included in the section, but I don't know what to write.

It should be okay now.

PR #3713, opened 2024-03-17
Update Sandman

love-code-yeyixiao:
Optimized a pair of conflict options in the Sandman sandbox options.

Added the ability and option to automatically lock the encrypted sandbox when it is closed.
And the feature on the title.


SandboxiePlus/SandMan/Windows/OptionsAdvanced.cpp
```diff
@@ -265,7 +266,13 @@ void COptionsWindow::LoadAdvanced()
 	ui.chkNotifyProtect->setChecked(m_pBox->GetBool("NotifyBoxProtected", false));
 
 	ui.chkProtectWindow->setChecked(m_pBox->GetBool("IsProtectScreen"));
-	ui.chkBlockCapture->setChecked(m_pBox->GetBool("IsBlockCapture"));
+	QString str = m_pBox->GetText("OpenWindowClass", "");
```
offhub, line 269:
```suggestion
	QString str = m_pBox->GetText("OpenWinClass", "");
```

DavidXanatos:
I dont think we need an explicit setting "LockWhenClose" as this unless the box is mounted manually already is the default, also using sandman for this task is not reliable as sandman may be closed.
This setting should be removed again instead we can add a chackbox to the mounting window allowing to specify the on box close behavioure which tehen is passed to the service to ensure reliable aplication.

PR #3714, opened 2024-03-17
Update OptionsWindow.ui

APMichael:
Shouldn't the previous description text be used again, as it is in the changelog?

### DIFF
--- a/SandboxiePlus/SandMan/Forms/OptionsWindow.ui
+++ b/SandboxiePlus/SandMan/Forms/OptionsWindow.ui
@@ -824,7 +824,7 @@
                <item>
                 <widget class="QCheckBox" name="chkBlockCapture">
                  <property name="text">
-                  <string>Block obtain an image of an un-sandboxied window through Windows public method</string>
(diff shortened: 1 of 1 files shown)


love-code-yeyixiao:
This is too one-sided.I suggest keeping the original explanation.

love-code-yeyixiao:
It functions as it was originally explained.

offhub:
If this setting does not distinguish whether applications/windows are in the sandbox or not, I think both sentences are incorrect.

APMichael:
@love-code-yeyixiao To be honest, I personally don't think that a normal user will understand your description text. David's text, on the other hand, is very understandable.

Why did you correct the changelog (#3715) before a decision was made here? Now it doesn't match again!

---

@offhub According to the previous texts, it refers to all windows outside the running sandbox, just as David had written. If this is not the case, then the text should be changed again.

---

@DavidXanatos Can you please make a decision so that the description text is the same both in the **OptionsWindow.ui** and in the **changelog**. Thank you very much!

PR #3715, opened 2024-03-17
Update CHANGELOG.md

love-code-yeyixiao:
Thank you for your contribution to the Sandboxie repository.

In order to reduce the risk of accidental merges, it's highly recommended for beginners to open a new Pull Request in draft state by clicking the button "Create Draft Pull Request", instead of using the default "Create Pull Request" option. 

In addition, you can convert an existing pull request to a draft by clicking the "Convert to draft" link in the right sidebar under "Reviewers".

All new translators are encouraged to look at the "Localization notes and tips" before creating a pull request: https://git.io/J9G19


bastik-1001:
Now the sentence seems to be wrong.

"Block obtain an image of an un-sandboxied window through Windows public method" seems to be missing something.

"Block to obtain an image of an un-sandboxed window through Windows public method" might work, but what does that even mean? What is "Windows public method" trying to say?

"Block taking screenshots of windows not belonging to the containing sandbox" is easy to understand, unless it can't be used as there are methods that still allow screenshots to be taken.

PR #3767, opened 2024-03-28
Update sandman_de.ts

bastik-1001:
Fresh fork to translate the new strings, oldsource removed, vanished entries removed.

The longest new string was practically already translated by APMichael, so I kept it close to it, but changed the wording to reflect the changes.


bastik-1001:
Michael was faster.

PR #3873, opened 2024-05-04
Force explorer child process

love-code-yeyixiao:
The purpose of this submission is to lay the groundwork for the “Sandbox desktop” feature that automatically opens each File Explorer's child processes into the corresponding sandbox.


offhub:
This causes a boot loop when you run a program from explorer.exe.

`SYSTEM_SERVICE_EXCEPTION (3b)`

PR #3952, opened 2024-05-27
Update sandman_de.ts

bastik-1001:
This adds translations for the new strings.

### DIFF
--- a/SandboxiePlus/SandMan/sandman_de.ts
+++ b/SandboxiePlus/SandMan/sandman_de.ts
@@ -5693,7 +5693,7 @@ This file is part of Sandboxie and all change done to it will be reverted next t
     <message>
         <location filename="Wizards/SetupWizard.cpp" line="539"/>
         <source>When this option is set, Sandbox Manager with normal user permissions will not be able to modify the configuration, which may result in a lock. You need to open the Sandbox Manager main window, click &quot;Sandbox (s)&quot; in the system menu, and then click &quot;Restart as Admin&quot; in the pop - up context menu to gain control of the configuration.</source>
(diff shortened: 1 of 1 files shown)


SandboxiePlus/SandMan/sandman_de.ts
```diff
@@ -5683,17 +5683,17 @@ This file is part of Sandboxie and all change done to it will be reverted next t
...
         <location filename="Wizards/SetupWizard.cpp" line="539"/>
         <source>Warning</source>
-        <translation type="unfinished"></translation>
+        <translation>Warnung</translation>
     </message>
     <message>
         <location filename="Wizards/SetupWizard.cpp" line="539"/>
         <source>When this option is set, Sandbox Manager with normal user permissions will not be able to modify the configuration, which may result in a lock. You need to open the Sandbox Manager main window, click &quot;Sandbox (s)&quot; in the system menu, and then click &quot;Restart as Admin&quot; in the pop - up context menu to gain control of the configuration.</source>
-        <translation type="unfinished"></translation>
+        <translation>Wenn diese Option gesetzt ist, ist es dem Sandboxverwalter mit normalen Nutzerrechten unmöglich die Konfiguration zu ändern, was dazu führen kann, dass Sie sich selbst sperren. Sie müssen das Hauptfenster des Sandboxverwalters öffnen und im Menü &quot;Sandbox (s)&quot; &quot;Als Admin neu starten&quot; auswählen, um die Kontrolle über die Konfiguration zu erhalten.</translation>
```
APMichael, line 5696: I would suggest not to translate “Manager” (as before) or to use the shortened “SandMan.

bastik-1001: I used the word manager. I decided against SandMan as the source does not contain the short version, either. Ultimately it will be SandMan, although sandbox manager is broad enough to include other tools.

APMichael: Fits, "Sandboxmanager" has also been used several times in the translations.

But I have just noticed that you have overlooked the second "Sandboxverwalters".

bastik-1001: You are right, I forgot to read the whole string again, and just replaced the first occurrence of it. Now I changed that, but used SandMan, since the line gives an instruction that only applies to SandMan.

PR #3991, opened 2024-06-14
Update NewBoxWizard

offhub:
- added checkbox for PromptForInternetAccess option
- improved boxToken checkbox


DavidXanatos:
Is there a reason we use InsertText instead of AppendText in the wizard?
I woudl liek to remove the use of InsertText from sandman entierly.

offhub:
> Is there a reason we use InsertText instead of AppendText in the wizard?

As far as I remember, there was no particular reason for it.

>  I woudl liek to remove the use of InsertText from sandman entierly.

Will this change affect the `insert` option of SbieIni.exe?

DavidXanatos:
> Will this change affect the insert option of SbieIni.exe?

No I would leave it in there for posterity, although in some new scenarios this may default to do an append instead

Issue #3999, opened 2024-06-18
User filtering


1mm0rt41PC:
### Describe what you noticed and did

I wanted the administrator account to bypass the sandbox.
Todo that,  I have created a group `grpSandboxed` and added to the ini configuration `Enabled=y,grpSandboxed`
But when I want to modify this box later, it is hidden and when I force the modification, a message indicate `This sandbox is disabled, do you want to enable it?`, the restriction on the "grpSandboxed" group is removed (ini is changed to `Enabled=y`) and all users are again targeted by Sandboxie.

### How often did you encounter it so far?

Always

### Expected behavior

The edition of the configuration should be allowed by the local admin without removing the filtering.

### Affected program

Not relevant

### Download link

Not relevant

### Where is the program located?

Not relevant to my request.

### Did the program or any related process close unexpectedly?

No, not at all.

### What version of Sandboxie are you running now?

1.15.0 (Insider Builds Sandboxie-Plus-Mar-16-2024.exe)

### Is it a new installation of Sandboxie?

I recently did a new clean installation.

### Is it a regression from previous versions?

No other version tested

### In which sandbox type you have this problem?

In a standard isolation sandbox (yellow sandbox icon).

### Can you reproduce this problem on a new empty sandbox?

I can confirm it also on a new empty sandbox.

### What is your Windows edition and version?

Windows Server 2022 & Windows 10 22H2

### In which Windows account you have this problem?

A local account (Standard user)., A local account (Administrator)., A Microsoft account (Standard user)., A Microsoft account (Administrator)., I use the built-in Administrator account.

### Please mention any installed security software

None & Windows Defender is disabled for these tests

### Did you previously enable some security policy settings outside Sandboxie?

Clean & refresh install of Windows Server 2022 & Win 10 22H2 - No custom config with the exception of Windows Defender that is disabled

### Sandboxie.ini configuration

```ini
[DefaultBox]
Enabled=y,grpSandboxed
BorderColor=#00ffff,ttl,6
Template=OpenBluetooth
Template=SkipHook
Template=FileCopy
Template=qWave
Template=BlockPorts
Template=LingerPrograms
Template=AutoRecoverIgnore
ConfigLevel=10
```

NixonSchool:
I can't make that long report that's too much. Fix the latest 20mb version x64-v1.13.7 - its not getting admin, yet I don't have UAC on win10. Its stuck at admin screen. 
[...]

PR #4028, opened 2024-06-24
Fix issue https://github.com/sandboxie-plus/Sandboxie/issues/3999

1mm0rt41PC:
Change message and behaviour to request edition instead of changing Enabled status:

`This sandbox is disabled, do you want to enable it?`
to
`This sandbox is disabled or restricted to a group/user, do you want to edit it?`

All translations have been updated


APMichael:
Please note: You only need to correct the string in the source file, but not in the individual language files. The language files are automatically updated after merging.

By the way, the German translation does not use the word "Sandkasten", but has always used "Sandbox".

APMichael:
> And what do we do now? the other transaltions may also be not proper?

Just a suggestion: If you agree with the change of the string, maybe you could make the changes in the source file "SbieView.cpp" and just close this pull request, right?

PR #4042, opened 2024-07-02
CrackKernelAndUserIntferace

love-code-yeyixiao:
The main purpose of this request is to test the feasibility of the jailbreaking measure and obtain its build results without the need for a merge.


love-code-yeyixiao:
So it's only a joy.

PR #4062, opened 2024-07-07
Add a second-auth feature in SbieStart

love-code-yeyixiao:
The purpose of this pull request is to provide an option to allow Start.exe to not just use the system's own UAC popup when attempting to start a programme that requires administrator privileges, 
but at the same time ask for the administrator account's credentials (which at this stage is also the password) again by us.


offhub:
Does this implementation support Windows Hello (like PIN)?

https://learn.microsoft.com/en-us/windows/win32/api/wincred/nf-wincred-creduipromptforwindowscredentialsw

love-code-yeyixiao:
If we use [creduipromptforwindowscredentialsw](https://learn.microsoft.com/en-us/windows/win32/api/wincred/nf-wincred-creduipromptforwindowscredentialsw),then we can do that,but it seems that we won't be able to vertify the truth of input,because we can't call LogonUser with a PIN blob,and there is no funcation like [CredUiConifrom](https://learn.microsoft.com/zh-cn/windows/desktop/api/wincred/nf-wincred-creduiconfirmcredentialsa) for [creduipromptforwindowscredentialsw](https://learn.microsoft.com/en-us/windows/win32/api/wincred/nf-wincred-creduipromptforwindowscredentialsw).So we have to use [creduipromptforcredentialsa](https://learn.microsoft.com/zh-cn/windows/win32/api/wincred/nf-wincred-creduipromptforcredentialsa)

love-code-yeyixiao:
Please review it. @DavidXanatos

PR #4114, opened 2024-07-24
Update sandman_de.ts

APMichael:
- Added new translations.
- Removed "vanished" messages.


bastik-1001:
"Create a new sandboxed token instead of stripping down the original token" turns to "Erzeuge einen neuen sandgeboxten Token, anstatt das ursprüngliche Token zu entfernen" how about "Erzeuge einen neuen sandgeboxten Token, anstatt das ursprüngliche Token zu beschränken"? or "einzuschränken" or something like that, as "der" or "das" Token does not get removed.

APMichael:
Yes, I also think that fits better. We should correct that in the next update:

"Erzeuge einen neuen sandgeboxten Token, anstatt den ursprünglichen Token zu beschränken"

(I couldn't really find a suitable translation for “stripping down” yesterday, so I had asked a translation service).

PR #4119, opened 2024-07-25
Fixed a few typos

APMichael:
- Fixed a few typos.


APMichael:
By the way, is the bolded part of the following sentence really correct? At least it might not be very understandable, right?

OptionsWindow.ui, 4799:
Some programs read system details through WMI (a Windows built-in database) instead of normal ways. For example, "tasklist.exe" could get full processes list even if "HideOtherBoxes" **is opened through accessing WMI**. Enable this option to stop this behaviour.

isaak654:
Good question. Perhaps one among @love-code-yeyixiao, @offhub, @diversenok or @DavidXanatos can clarify your doubt.

diversenok:
`HideOtherBoxes` hides information from syscall-based process enumeration APIs; it doesn't cover WMI. So tools that use WMI can still naturally see processes from other boxes. You can use `BlockAccessWMI` to deny access to WMI altogether to cover this scenario. But keep in mind that even without WMI, there are still (not widely known/used) ways to enumerate all processes on the system and Sandboxie cannot do anything about them. So don't treat `HideOtherBoxes` as a security boundary.

APMichael:
Thank you very much for explaining the options.

My question was more about whether the tooltip description is really correct and understandable. The description used previously was simpler: "...even if 'HideOtherBoxes' **is enabled.**"

love-code-yeyixiao:
The sentence means "tasklist.exe" could get full process list,when it uses WMI,even though the option "HideOtherBoxes" is opened.

love-code-yeyixiao:
> Thank you very much for explaining the options.
> 
> My question was more about whether the tooltip description is really correct and understandable. The description used previously was simpler: "...even if 'HideOtherBoxes' **is enabled.**"

This sentence does not reflect the tasklist's use of WMI.

wilders-soccerfan:
OptionsWindow.ui, line 4296:
Is the space before iexplore.exe intentional in 'ClosedFilePath=! iexplore.exe,C:Users*' ?

APMichael:
@love-code-yeyixiao Thank you for your reply.

I think we have a misunderstanding. The whole sentence already says that a complete list of processes can be retrieved via WMI, for example. I was only concerned with the part marked in bold "...is opened through accessing WMI.", which in my opinion is not understandable. This would make more sense to me, for example:

"...For example, "tasklist.exe" could get full processes list through accessing WMI, even if "HideOtherBoxes" is enabled."

APMichael:
@wilders-soccerfan Thanks for the hint.

I didn't even notice that because I only corrected the apostrophe characters. But I also think that the space is unnecessary.

SandboxiePlus/SandMan/Forms/OptionsWindow.ui
```diff
@@ -4796,10 +4796,10 @@ This is done to prevent rogue processes inside the sandbox from creating a renam
              <item row="12" column="0" colspan="5">
               <widget class="QCheckBox" name="chkBlockWMI">
                <property name="toolTip">
-                <string>Some programs read system deatils through WMI(A Windows built-in database) instead of normal ways. For example,&quot;tasklist.exe&quot; could get full processes list even if &quot;HideOtherBoxes&quot; is opened through accessing WMI. Enable this option to stop these behaviour.</string>
+                <string>Some programs read system details through WMI (a Windows built-in database) instead of normal ways. For example, &quot;tasklist.exe&quot; could get full processes list even if &quot;HideOtherBoxes&quot; is opened through accessing WMI. Enable this option to stop this behaviour.</string>
```
isaak654, line 4799:
```suggestion
                <string>Some programs read system details through WMI (a Windows built-in database) instead of normal ways. For example, &quot;tasklist.exe&quot; could get full processes list through accessing WMI, even if &quot;HideOtherBoxes&quot; is used. Enable this option to stop this behaviour.</string>
```

SandboxiePlus/SandMan/Forms/OptionsWindow.ui
```diff
@@ -4796,10 +4796,10 @@ This is done to prevent rogue processes inside the sandbox from creating a renam
...
                </property>
                <property name="text">
-                <string>Prevent sandboxed processes from accessing system deatils through WMI (see tooltip for more Info)</string>
+                <string>Prevent sandboxed processes from accessing system details through WMI (see tooltip for more Info)</string>
```
isaak654, line 4802:
```suggestion
                <string>Prevent sandboxed processes from accessing system details through WMI (see tooltip for more info)</string>
```

SandboxiePlus/SandMan/Forms/OptionsWindow.ui
APMichael: Thank you. I hope that everyone agrees with this.

PR #4126, opened 2024-07-28
DiskSerialNumber Spoof And ProductId Spoof

love-code-yeyixiao:
This pull request include the following purpose:

1. Implement a feature that randomly modifies the Windows product ID in the registry when the first application in the sandbox is started.
2. Implement a function that returns a random value when the application tries to get the partition serial number.


DavidXanatos:
for anonymity "SOFTWARE\Microsoft\Cryptography" "MachineGuid" should also be altered

love-code-yeyixiao:
> for anonymity HKLM "SOFTWARE\Microsoft\Cryptography" "MachineGuid" should also be altered

HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\SQMClient\MachineId
And this?

PR #4140, opened 2024-08-05
AlwaysActive: force active-window state for sandboxed apps

love-code-yeyixiao:
The purpose of this submission is to allow sandboxed programs to always consider their windows to be the system's active window, thus indirectly permitting some games to run in the background.


Sandboxie/core/dll/gui.c
```diff
@@ -1668,6 +1668,19 @@ _FX LRESULT Gui_WindowProcW(
...
+	if (uMsg == WM_KILLFOCUS) {
+		if (SbieApi_QueryConfBool(NULL, L"AlwaysActive", FALSE))
+			return NULL;
+	}
+	if (uMsg == WM_ACTIVATE) {
+		if (SbieApi_QueryConfBool(NULL, L"AlwaysActive", FALSE))
+			if (wParam == WA_INACTIVE)
+				return 0;
+			else {
+				Gui_PervousActivedWindow = (HWND)hWnd;
```
offhub, line 1680:
```suggestion
				Gui_PreviousActiveWindow = (HWND)hWnd;
```

Sandboxie/core/dll/gui_p.h
```diff
@@ -519,7 +521,7 @@ extern WCHAR* Gui_BoxNameTitleW;
 extern ATOM Gui_DropTargetProp_Atom;
 extern ATOM Gui_WindowProcOldW_Atom;
 extern ATOM Gui_WindowProcOldA_Atom;
-
+extern HWND Gui_PervousActivedWindow = NULL;
```
offhub, line 524:
```suggestion
extern HWND Gui_PreviousActiveWindow = NULL;
```

Sandboxie/core/dll/guimisc.c
```diff
@@ -1707,3 +1716,18 @@ _FX void Gui_SwitchToThisWindow(HWND hWnd, BOOL fAlt)
 		return;
 	__sys_SwitchToThisWindow(hWnd, fAlt);
 }
+
+//---------------------------------------------------------------------------
+//Gui_GetActiveiWindow
+//---------------------------------------------------------------------------
+static HWND Gui_GetActiveWindow() {
+	if (SbieApi_QueryConfBool(NULL, L"AlwaysActive", FALSE))
+		return (Gui_PervousActivedWindow == NULL ? __sys_GetActiveWindow() : Gui_PervousActivedWindow);
```
offhub, line 1725:
```suggestion
		return (Gui_PreviousActiveWindow == NULL ? __sys_GetActiveWindow() : Gui_PreviousActiveWindow);
```

Sandboxie/core/dll/guimisc.c
```diff
@@ -1707,3 +1716,18 @@ _FX void Gui_SwitchToThisWindow(HWND hWnd, BOOL fAlt)
...
+	return __sys_GetActiveWindow();
+}
+
+static HWND Gui_GetForegroundWindow() {
+	if (SbieApi_QueryConfBool(NULL, L"AlwaysActive", FALSE))
+		return (Gui_PervousActivedWindow == NULL ? __sys_GetActiveWindow() : Gui_PervousActivedWindow);
```
offhub, line 1731:
```suggestion
		return (Gui_PreviousActiveWindow == NULL ? __sys_GetActiveWindow() : Gui_PreviousActiveWindow);
```

leonardo3dp:
Would there be a way to tell the application inside the sandbox that the mouse is always over the window?

leonardo3dp:
I couldn't make it work
SBIE2303 Could not hook GetActiveWindow (1)
SBIE2318 DLL initialization failed for 'user32.dll'

love-code-yeyixiao:
Found a problem, I'll deal with it later.












YeYixiao
***@***.***






---Original---
From: ***@***.***&gt;
Date: Fri, Jul 31, 2026 23:52 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [sandboxie-plus/Sandboxie] AlwaysActive (PR #4140)




 
@Copilot commented on this pull request.
 
🟡 Not ready to approve
 
The current AlwaysActive WM_ACTIVATE suppression and HWND tracking can break correct activation semantics (especially intra-process/multi-thread UI) and should be tightened/validated before approval.
 
Once you've addressed the issues Copilot identified, you can request another Copilot review.
 
This review doesn't count toward merge requirements. Sign up for the private preview to control whether Copilot approvals count.
  Review details 
Suppressed comments (5)
 
Sandboxie/core/dll/gui.c:1768
  
Same issue as the WndProcW path: suppressing WA_INACTIVE for WM_ACTIVATE unconditionally can break normal intra-process window activation (lParam is the HWND being activated). Consider suppressing deactivation only when the new active window is outside the current process, and remove the commented-out WM_KILLFOCUS block.
  	//if (uMsg == WM_KILLFOCUS) { 	//	if (SbieApi_QueryConfBool(NULL, L"AlwaysActive", FALSE)) 	//		return FALSE; 	//} 	if (uMsg == WM_ACTIVATE &amp;&amp; Gui_AlwaysActive) { 		switch (LOWORD(wParam)) { 		case WA_INACTIVE: 			return 0; 		case WA_ACTIVE: 		case WA_CLICKACTIVE: 			Gui_PreviousActiveWindow = hWnd; 			break; 		} 	}  
Sandboxie/core/dll/gui.c:1689
  
The WM_ACTIVATE handler suppresses WA_INACTIVE unconditionally, which can leave multiple windows in the same process thinking they are active when switching between them (lParam points at the window being activated). Consider only suppressing deactivation when activation is moving to a window outside the current process, and remove the commented-out WM_KILLFOCUS block.
  
This issue also appears on line 1755 of the same file.
 	//if (uMsg == WM_KILLFOCUS) { 	//	if (SbieApi_QueryConfBool(NULL, L"AlwaysActive", FALSE)) 	//		return FALSE; 	//} 	if (uMsg == WM_ACTIVATE &amp;&amp; Gui_AlwaysActive) { 		switch (LOWORD(wParam)) { 		case WA_INACTIVE: 			return 0; 		case WA_ACTIVE: 		case WA_CLICKACTIVE: 			Gui_PreviousActiveWindow = hWnd; 			break; 		} 	}  
Sandboxie/core/dll/guimisc.c:1760
  
GetActiveWindow is defined as per-thread (active window for the calling thread’s message queue), but AlwaysActive tracks a single process-global Gui_PreviousActiveWindow that may be updated from a different UI thread. This can return the wrong HWND in multi-UI-thread apps; consider storing the previous active HWND in TLS/THREAD_DATA (and/or keying by idEventThread for WinEvent filtering).
  HWND Gui_PreviousActiveWindow = NULL; static HWND Gui_GetActiveWindow(void) { 	if (Gui_AlwaysActive &amp;&amp; Gui_PreviousActiveWindow) 		return Gui_PreviousActiveWindow; 	return __sys_GetActiveWindow();  
Sandboxie/core/dll/guimisc.c:1768
  
Gui_GetActiveWindow / Gui_GetForegroundWindow return Gui_PreviousActiveWindow without validating that the HWND is still valid. If the handle becomes stale (e.g., missed WM_NCDESTROY path), callers will get an invalid HWND; consider guarding with __sys_IsWindow (and falling back to the real API).
  static HWND Gui_GetActiveWindow(void) { 	if (Gui_AlwaysActive &amp;&amp; Gui_PreviousActiveWindow) 		return Gui_PreviousActiveWindow; 	return __sys_GetActiveWindow(); } static HWND Gui_GetForegroundWindow(void) { 	if (Gui_AlwaysActive &amp;&amp; Gui_PreviousActiveWindow) 		return Gui_PreviousActiveWindow; 	return __sys_GetForegroundWindow(); }  
Sandboxie/core/dll/gui_p.h:521
  
Gui_PreviousActiveWindow is shared across gui.c and guimisc.c but is declared as a local extern inside gui.c. Declaring it in gui_p.h (next to Gui_AlwaysActive) avoids hidden cross-file coupling and makes the interface explicit.
  extern BOOLEAN Gui_AlwaysActive;   
Files reviewed: 3/3 changed files
 
Comments generated: 0 new
 
Review effort level: Lite
   
We're testing this review assessment. Please use 👍 or 👎 to tell us if it's correct.
 
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS and Android. Download it today!

love-code-yeyixiao:
Currently, the game Plants vs. Zombies can still detect changes to the active window.

Fix: Now it works well.

love-code-yeyixiao:
> Would there be a way to tell the application inside the sandbox that the mouse is always over the window?

It might be possible to map cursor pos and movement with RawInput, but at least not in this PR.

love-code-yeyixiao:
Due to the special nature of sbiedll.dll, performing too many unload operations offers no benefit. When a process terminates, kernel objects such as hooks are reclaimed by the operating system kernel. Therefore, the suppressed suggestions from this Copilot are unnecessary—because under normal circumstances, they are simply not needed. There might be an extremely remote race condition issue, but that is inconsequential when the process is already destined to terminate.

Kizuno18:
I reviewed the current head (`8471104ff4823f158685d1f9eaee0b10aa433f9f`) and found a reproducible multi-UI-thread state bug in the AlwaysActive message handling.

When activation or focus moves from window A to window B on another UI thread in the same process, `Gui_AlwaysActiveMsg` correctly lets `WA_INACTIVE` / `WM_KILLFOCUS` pass through, but it leaves A in that thread's `gui_active_window` / `gui_focus_window` TLS cache. Because A remains a valid HWND, later `GetActiveWindow`, `GetFocus`, and `GetGUIThreadInfo` calls from the losing thread can still report A. The `WM_NCACTIVATE(FALSE)` path also suppresses A's nonclient deactivation without checking its documented `lParam` target, so an ordinary same-process switch can leave both captions looking active.

I prepared a narrow fix on top of this exact head:

- commit: https://github.com/Kizuno18/sbxie/commit/19c07b8e767d0f29c581608334b08455ddb056dc
- branch comparison: https://github.com/sandboxie-plus/Sandboxie/compare/AlwaysActive...Kizuno18:sbxie:codex/help-pr-4140

The fix clears only the losing thread's TLS entries for same-process transfers, uses the `WM_NCACTIVATE` target process to preserve normal internal caption changes, and returns the real foreground/thread information while the foreground still belongs to the process. External deactivation remains suppressed and keeps the cached state.

Validation: a deterministic MSVC `/W4 /WX` handler harness passed 10 assertions covering same-process and external activation, focus, and nonclient transitions. The Win32 `SandboxDll.sln` build compiled both changed production sources cleanly; the host then stopped at the existing resource dependency because `afxres.h` is unavailable in this Build Tools installation.

Follow-up: during the final integration review, I also removed the newly added `Gui_UninitMisc` teardown and its bookkeeping flag because `Gui_UninitMisc` has no caller and is not exported. This keeps the runtime hook behavior unchanged and avoids carrying a dead cleanup path: https://github.com/Kizuno18/sbxie/commit/c01ce8fa

PR #4200, opened 2024-09-01
ZipExportImport

love-code-yeyixiao:
Allow users to export/import boxes with .zip files.


DavidXanatos:
I have fixed the underlaying 7z issue you can add the UI changes and it should work how

Issue #14, opened 2020-04-16
Malware and fake dev


burnsjackie:
official Sandboxie found here: https://github.com/sandboxie/sandboxie

this one is not official, exe signed with malware signer this one. avoid

VitRom:
hey, shiteater, what about proof?
don't you know such a loud words need some confirmation, arshead?

> 
> 
> official Sandboxie found here: https://github.com/sandboxie/sandboxie
> 
> this one is not official, exe signed with malware signer this one. avoid

Justsnoopy30:
The official sandboxie is no longer being maintained, so now the community is trying to maintain it.

VitRom:
Wow! Such a fresh news! I'm so glad to get know this, thank you very much.

And now let's come back to the second part of your first poop:
What. About. Proof. Qestionmark.

VitRom:
> 
> 
> The official sandboxie is no longer being maintained, so now the community is trying to maintain it.

Good morning, dear "community that is trying to maintain". So nice you're awoken. Finally.
But https://github.com/sandboxie-plus/Sandboxie/network vs https://github.com/sandboxie/sandboxie/network shows clear who are trying and who are talking.

Justsnoopy30:
I was not agreeing with the person who opened the issue if you couldn't tell.

PR #4218, opened 2024-09-07
fix get_openssl.cmd

pedoc:
Due to the change of the target download address, the synchronization is changed.

Latest download address:
``https://github.com/xanasoft/openssl-builds/releases/download/openssl-3.3.1/openssl-3.3.1.zip``


offhub:
I don't see any problems that need fixing. See line [#14](https://github.com/sandboxie-plus/Sandboxie/blob/master/Installer/get_openssl.cmd#L14)

---

To clarify, this command:

```batch
set "openssl_version_underscore=%openssl_version:.=_%"
```

will take the value of `openssl_version`, replace all dots with underscores, and store the result in the `openssl_version_underscore` variable. 

For example, if `openssl_version` is `1.2.3`, then `openssl_version_underscore` will be `1_2_3`.

The `openssl-` part of the download address is not affected by this operation and is not part of the substitution logic here.

pedoc:
> I don't see any problems that need fixing. See line [#14](https://github.com/sandboxie-plus/Sandboxie/blob/master/Installer/get_openssl.cmd?rgh-link-date=2024-09-07T14%3A48%3A05Z#L14)

Sorry I didn't notice this line, the original script has already tried the correct download address here

Close this PR

PR #4222, opened 2024-09-08
CleanAfterReboot

Jkkoi:
The PR provides a menu action to clean FileRootPath of a box after the machine reboot.


DavidXanatos:
The code does not compile, see failed CI runs.
Further more, I'm not sure if a feature like this implemented this way is desirable. 
Not all sandboxes may be intended to be deleted, some may contain persistent software installations, etc.
This feature also bypasses the snapshot mechanism and just deletes all.
Imho it should be implemented differently allow per box granularity and the ability to roll back to last snapshot instead of deleting all.

SandboxiePlus/SandMan/Views/SbieView.cpp
```diff
@@ -1386,6 +1389,21 @@ void CSbieView::OnSandBoxAction(QAction* Action, const QList<CSandBoxPtr>& SandB
...
 	}
+	else if (Action == m_pRebootClean)
+    {
+		CSandBoxPtr pBox=SandBoxes.first();
+		auto pBoxEx = pBox.objectCast<CSandBoxPlus>();
+		QString fileRoot=pBoxEx->GetFileRoot();
+		if(QMessageBox::question(this, tr("Sandboxie-Plus"), 
+		tr("Do you want to make the box cleaned after machine reboot?"), 
+		QMessageBox::Yes, QMessageBox::No)==QMessageBox::Yes){
+			if(MoveFileExW(fileRoot.toStdString().c_str(),NULL,4)==0){
```
love-code-yeyixiao, line 1400: ExW->ExA

DavidXanatos:
```if(MoveFileExW(fileRoot.toStdWString().c_str(),NULL,4)==0){```
would be better as it would work with unicode paths, 
but that does not address my initial concerns, imho such a feature should be implemented in sandman in a way that basically triggers the normal delete box content code path for selected boxes after reboot.
and not using windows delayed file operations

**Comment by Jkkoi** [comment]
This will ensure that the deletion operation gets done reliably, especially if Sandman doesn't boot itself at boot time.

**Comment by Jkkoi** [comment]
@DavidXanatos   Please review it.❤️

File: SandboxiePlus/SandMan/Views/SbieView.cpp [hunk 2]
```diff
@@ -1386,6 +1389,27 @@ void CSbieView::OnSandBoxAction(QAction* Action, const QList<CSandBoxPtr>& SandB
...
+		CSandBoxPtr pBox=SandBoxes.first();
+		auto pBoxEx = pBox.objectCast<CSandBoxPlus>();
+		QString fileRoot=pBoxEx->GetFileRoot();
+		if(QMessageBox::question(this, tr("Sandboxie-Plus"), 
+		tr("Do you want to make the box cleaned after machine reboot?"), 
+		QMessageBox::Yes, QMessageBox::No)==QMessageBox::Yes){
+            if(QMessageBox::question(this, tr("Sandboxie-Plus"),
+            tr("Do you want to use Windows Delecting Feature(MoveFileEx)?Or we'll delete the box when Sandman.exe called by userinit.exe."),
+                QMessageBox::Yes, QMessageBox::No)==QMessageBox::Yes){
+                if(MoveFileExW(fileRoot.toStdString().c_str(),NULL,4)==0){
```
love-code-yeyixiao, line 1403: MoveFileExW->MoveFileExA

Jkkoi:
@DavidXanatos I have provided another clean option.

offhub:
It might be better to check for a 'DON'T-USE.TXT' file in both the sandbox directory and its parent directory. If the files are not found, the operation (MoveFileEx) should be aborted.


sbiePR4222-dontuse

SandboxiePlus/SandMan/Views/SbieView.cpp
```diff
@@ -1386,6 +1389,28 @@ void CSbieView::OnSandBoxAction(QAction* Action, const QList<CSandBoxPtr>& SandB
 
 		Results.append(Status);
 	}
+	else if (Action == m_pRebootClean)
+    {
+		CSandBoxPtr pBox=SandBoxes.first();
+		auto pBoxEx = pBox.objectCast<CSandBoxPlus>();
+		QString fileRoot=pBoxEx->GetFileRoot();
+		if(QMessageBox::question(this, tr("Sandboxie-Plus"), 
+		tr("Do you want to make the box cleaned after machine reboot?"),
```
offhub, line 1398:
```suggestion
		tr("Do you want the box to be cleaned after the machine reboots?"), 
```

SandboxiePlus/SandMan/Views/SbieView.cpp
```diff
@@ -1386,6 +1389,28 @@ void CSbieView::OnSandBoxAction(QAction* Action, const QList<CSandBoxPtr>& SandB
...
+	else if (Action == m_pRebootClean)
+    {
+		CSandBoxPtr pBox=SandBoxes.first();
+		auto pBoxEx = pBox.objectCast<CSandBoxPlus>();
+		QString fileRoot=pBoxEx->GetFileRoot();
+		if(QMessageBox::question(this, tr("Sandboxie-Plus"), 
+		tr("Do you want to make the box cleaned after machine reboot?"), 
+		QMessageBox::Yes, QMessageBox::No)==QMessageBox::Yes){
+            if(QMessageBox::question(this, tr("Sandboxie-Plus"),
+            tr("Do you want to use Windows Delecting Feature(MoveFileEx)?Or we'll delete the box when Sandman.exe called by userinit.exe."),
```
offhub, line 1401:
```suggestion
            tr("Do you want to use the Windows deleting function (MoveFileEx), or delete the box when SandMan.exe starts?"),
```

SandboxiePlus/SandMan/Views/SbieView.cpp
```diff
@@ -1386,6 +1389,28 @@ void CSbieView::OnSandBoxAction(QAction* Action, const QList<CSandBoxPtr>& SandB
...
+                QMessageBox::Yes, QMessageBox::No)==QMessageBox::Yes){
+                if(MoveFileExA(fileRoot.toStdString().c_str(),NULL,4)==0){
+                    QMessageBox::warning(this, tr("Sandboxie-Plus"),
+                    tr("The operation failed,please make sure that Sandman has admin privliage.")
```
offhub, line 1405:
```suggestion
                    tr("The operation failed. Please make sure that SandMan has administrative privileges.")
```

offhub (changes requested):

DavidXanatos:
Did you test the behavior if MoveFileEx when the to be deleted sandbox contains symlinkd and or directory junctions.
Also with the new file permission scheme used in 1.15.x we in particular 'UseOriginalACLs=y' MoveFileEx may not have enough permissions to delete the folder.

Hence executive decision:
We will not use MoveFileEx to delete files on reboot, at all.

In an upcoming build of Sandboxie we well add a mechanism to perform the deletion by sbiesvc with system privileges, for that we need to ensure that we don't open a backdoor allowing anyone to delete any folder on the system willy-nilly, checking for the presence of DON'T-USE.TXT or a sbie desktop.ini is a good start, although I think ultimately we will use the ACLs of the containing folder that is if the user has permissions to delete the parent folder we grant these permissions to all sub folders even if inheritance is disabled and that permission is not set for a sub folder or file.

If we desire to implement a delete after reboot feature we can tie this into that new service functionality.
Therefor I'll close this pull request now.

@Jkkoi You are very welcome to help with the new service based approach.

PR #4250, opened 2024-09-17
Update sandman_de.ts

APMichael:
- Added new translations.
- Checked translations with modified source.
- Removed "vanished" message.

### DIFF
--- a/SandboxiePlus/SandMan/sandman_de.ts
+++ b/SandboxiePlus/SandMan/sandman_de.ts
@@ -822,12 +822,12 @@ Klicken Sie auf Abschließen um den Assistenten zu schließen.</translation>
     <message>
         <location filename="Windows/CompressDialog.cpp" line="27"/>
         <source>7-Zip</source>
-        <translation type="unfinished"></translation>
+        <translation>7-Zip</translation>
     </message>
     <message>
         <location filename="Windows/CompressDialog.cpp" line="28"/>
         <source>Zip</source>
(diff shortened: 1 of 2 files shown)


SandboxiePlus/SandMan/sandman_de.ts
```diff
@@ -6826,39 +6814,35 @@ Wenn Sie bereits ein Great Supporter auf Patreon sind, kann Sandboxie online nac
...
     <name>ExtractDialog</name>
     <message>
         <location filename="Forms/ExtractDialog.ui" line="14"/>
         <source>Extract Files</source>
-        <translation type="unfinished"></translation>
+        <translation>Dateien entpacken</translation>
     </message>
     <message>
         <location filename="Forms/ExtractDialog.ui" line="25"/>
         <source>Export Sandbox from an archive</source>
```
APMichael, line 6829: Shouldn't that actually be "**Import** Sandbox from an archive"?

bastik-1001: Either that, or it should be "Export Sandbox to an archive" (Sorry if that is resolved.)

APMichael: Thank you for your reply. Since it is the "Extract / Sandbox Import" dialog, I still think it should be "Import...".

bastik-1001: From the logical point of view, this should indeed be the case.

PR #4268, opened 2024-09-27
Fix buffer overflow in GetRawInputDeviceInfo proxy

### LINKED ISSUE #4267: Proxy for GetRawInputDeviceInfoW() causes a buffer overflow

See Firefox [Bug 1921198](https://bugzilla.mozilla.org/show_bug.cgi?id=1921198#c1) - Crash in [@ memcpy | sbiedll.dll | GetRawInputDeviceInfo] caused by sandboxie

It would likely happen to any user that is using a gamepad on Firefox in Sandboxie.

### Expected behavior

Sandboxie's implementation needs to be changed so that it only considers the `pcbSize` argument to be wide-chars when `uiCommand == RIDI_DEVICENAME`.

### Affected program

Firefox

### Download link

https://mozilla.org

Not relevant to my request.

Yes, it did, but no .dmp file has been created in the system.

### Crash dump

_No response_

I'm not the one experiencing the crash, but it should happen on the lastest release.

¯\\\_(ツ)_/¯

Not relevant to my request.

All versions of Windows since Windows XP

Not relevant to my request.

N/A

_No response_

### Trace log

_No response_

_No response_

marti4d:
The proxy for GetRawInputDeviceInfo() incorrectly assumes that pcbSize is always in wchars when unicode is used, which is only true if the name of the device is being queried (`uiCommand == RIDI_DEVICENAME`).

Otherwise, it is the exact size of the buffer in bytes. Right now, that means that the proxy will write double the number of bytes it's supposed to for the other `uiCommand` values, potentially causing a buffer overflow.

Also, when no `pData` value is given, `*pcbSize` is allowed to be uninitialized. The proxy currently reads it unconditionally as a UINT, which would be UB in that case. `memcpy` is able to safely copy uninitialized values, so let's use that instead!

Fixes #4267

### DIFF
--- a/Sandboxie/core/dll/guimisc.c
+++ b/Sandboxie/core/dll/guimisc.c
@@ -1330,10 +1330,13 @@ _FX LONG Gui_GetRawInputDeviceInfo_impl(
     GUI_GET_RAW_INPUT_DEVICE_INFO_REQ* req;
     GUI_GET_RAW_INPUT_DEVICE_INFO_RPL* rpl;
 
-    // Note: pcbSize seems to be in tchars not in bytes!
     ULONG lenData = 0;
-    if (pData && pcbSize)
-        lenData = (*pcbSize) * (bUnicode ? sizeof(WCHAR) : 1);
+    if (pData && pcbSize) {
+        lenData = *pcbSize;
+        if (uiCommand == RIDI_DEVICENAME && bUnicode) {
+            lenData *= sizeof(WCHAR);
+        }
+    }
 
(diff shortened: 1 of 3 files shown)


DavidXanatos:
Is the patch still work in progress or does it only missing some testing?

marti4d:
@DavidXanatos  It's only missing testing -- I just set it to "draft" so it would not be accidentally merged until someone has verified that it works.

DavidXanatos:
ok great I'll merge and debug it

DavidXanatos:
one issue i see already ULONG status; mist always be the first fiels in a reply structure even if it appears unused

PR #4284, opened 2024-10-05
Ability to hide certificate in the editbox

love-code-yeyixiao:
I Intolerance that the location of the certificates in the global settings was too easy to leak when the screen was shared, so I made the commit.


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [1.15.1 / 5.70.1] - 2024-10-
 
+### Added
+- added the ability to hide certificates in editbox in Global Setting
+
```
isaak654, line 11: Not under 1.15.1 / 5.70.1, which has already been released.

DavidXanatos:
What is to be protected the functionality of the cert or the name itself?
I would may be preffer to colals the edit box and expand it with a link as its done for the serial entry line edit.

DavidXanatos:
I added it my way making it behave like the serial entry: https://github.com/sandboxie-plus/Sandboxie/commit/04e9c0745cf936cd53527c82cbba15464c1207ce

PR #4300, opened 2024-10-16
Set tab orders and buddies of UI controls

gexgd0419:
This sets the tab orders for most dialogs, so that when the user presses Tab/Shift+Tab, or presses an arrow key on a button, the focus will move from left to right, and from top to bottom. This can make keyboard navigation easier.

Unfortunately, when different tabs are merged by code, their tab orders seem to be lost. The File Recovery and Stop Behaviour tabs still have incorrect tab orders.

This also sets the "buddy" for some labels, if the label contains a description for the following control. This can improve accessibility because screen readers can read the label for the user when entering a text box, for example.

When there are new controls added, their tab orders should be set as well.

### DIFF
--- a/SandboxiePlus/SandMan/Forms/BoxImageWindow.ui
+++ b/SandboxiePlus/SandMan/Forms/BoxImageWindow.ui
@@ -36,6 +36,9 @@
        <property name="alignment">
         <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
        </property>
+       <property name="buddy">
+        <cstring>txtRepeatPassword</cstring>
+       </property>
       </widget>
      </item>
      <item row="7" column="2">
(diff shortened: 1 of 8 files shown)


gexgd0419:
This should preserve the tab orders when merging the tabs, so that File Recovery and Stop Behaviour tabs can have the intended tab orders.

DavidXanatos:
How can I avoid messing up the tab order in future?

gexgd0419:
I changed the tab order manually with the Qt Creator, by clicking Edit menu > Edit Tab Order, then clicking everything in the desired order. But yes, this can be a hassle for complex dialogs.

So here in this PR #4310 I try to solve this issue by using code to set the tab order for all controls. It iterates over all controls, and sets their tab order according to their row and column number in the grid layout. If this can work, you won't need to manually adjust the tab layout.

Note that the PR assumes that the controls are always put in a layout (grid, form, horizontal or vertical), the only kind of widget that contains other widgets without using a layout is `QTabWidget`, and the correct tab order is always first from left to right (in the same row), then from top to bottom.

You can remove the existing `<tabstops>` element from the .ui files to test whether this will work without manually setting the tab order.

qqeee-ma:
How to make a donation?

PR #4310, opened 2024-10-19
Add helper for setting tab orders automatically

gexgd0419:
This PR adds a helper function that sets the tab order automatically using the layout information.

The helper functions are put into separate files (`TabOrder.h` and `TabOrder.cpp`) under the `Helpers` folder.

The helper is used in `OptionsWindow` and `SettingsWindow`.

The helper sets the tab order of a widget using its row and column number in the grid layout. The tab orders are assigned first from left to right (in the same row), then from top to bottom.

If in the future a different custom type of layout is used, or a container other than `QTabWidget` is used, then the helper functions may need to be changed.

Usage:

``` c++
#include "Helpers/TabOrder.h"

// Add the following line at the end of the constructor of a window:
SetTabOrder(this);
```

### DIFF
--- a/SandboxiePlus/SandMan/Helpers/TabOrder.cpp
+++ b/SandboxiePlus/SandMan/Helpers/TabOrder.cpp
@@ -1,0 +1,124 @@
+#include "stdafx.h"
+#include "TabOrder.h"
+
+// Items to be sorted by row & col
+template <class T>
+struct ChildItem
+{
+	T item;
+	int row, col;
+	template <class U>
+	ChildItem(U&& item, int row, int col)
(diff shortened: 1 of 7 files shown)


isaak654:
#4300

gexgd0419:
@isaak654 What was the thing here that referenced #4300? Force-pushing might have overwritten something...

isaak654:
> @isaak654 What was the thing here that referenced #4300? Force-pushing might have overwritten something...

It's me writing #4300 first, with the difference that I deleted my previous comment. In my opinion, it helps readability when an issue or pull remains pending.

PR #4321, opened 2024-10-23
Update sandman_zh_CN.ts

CybCom:
Fix wrong translations.
The original translations:
"重置密码" is Reset Password, not Repeat Password.
"加密密钥" is Key, not Cipher.


CybCom:
In addition, may I ask if I need to submit any application to obtain the contributor certificate? If it cannot be obtained, are there any specific criteria?

CybCom:
> In addition, may I ask if I need to submit any application to obtain the contributor certificate? If it cannot be obtained, are there any specific criteria?

Hi @DavidXanatos

PR #4322, opened 2024-10-23
Update sandman_de.ts

APMichael:
- Checked translation with modified source.
- Removed \<oldsource\> line.


APMichael:
> Is it necessary to remove the line manually? Won't the CI change it?

Yes, if you want to remove these lines, you have to do this manually. This also applies to messages that are marked as "vanished". Here you can delete the corresponding block `<message> ... </message>`.

However, deletion is not a must, you can also simply ignore these obsolete lines.

CybCom:
got it, thank you:)

PR #4370, opened 2024-11-16
Qt 5.15.16 with OpenSSL 3.4.0

offhub:


DavidXanatos:
Thanks I have updated to my builds What would be a great improvement would be to change the scripts such that the version number and hash only needs to be changed in one place not in 5 files.

offhub:
> Thanks I have updated to my builds What would be a great improvement would be to change the scripts such that the version number and hash only needs to be changed in one place not in 5 files.

#4376

PR #4376, opened 2024-11-17
Build variables are gathered in a single file

offhub:
except `qt6_version`  and `sslMajorVersion`.


DavidXanatos:
great will merge it after 1.15.2 is out
what would be also great would be a CI build run for an own Qt 6 build

PR #4584, opened 2025-03-16
SandboxPlus Project Improvement

pwnmelife:
fix an compilation error about SandboxPlus
change SandboxPlus result from \$(Platform) to Bin\\$(Platform) for the Installer\copy_build.cmd


DavidXanatos:
the idea was that for release we use the qt build script not the VS project as wen cant run the VS project on the CI infrastructure as it lacks the VS-Qt extension, but I suppose its reasonable to align the vs project build output with the qt script

PR #4613, opened 2025-03-28
Update sandman_zh_CN.ts

fzxx:
Fixed the translation, I also realised that the previous translation didn't translate (.) to (。)
Does it need to be amended substantially? @DavidXanatos @nkh0472


nkh0472:
> I also realised that the previous translation didn't translate (.) to (。)
> Does it need to be amended substantially?

The Chinese period "。" is more formal, and this change does not affect understanding.
The adjustment is reasonable. 

But be careful not to replace English periods in expressions or file extensions incorrectly.

fzxx:
Changed English symbols to Chinese symbols.

PR #4776, opened 2025-05-04
Alarm the potential escape risk

love-code-yeyixiao:
The existing prompt may encourage users to readily accept the privilege-escalation request of a program in the sandbox, but this can lead to a known but unpat sandbox escape issue.


Sandboxie/msgs/Sbie-English-1033.txt
```diff
@@ -1084,7 +1084,7 @@ User Account Control pop-up window.
 
 3243;txt;01
 Note:  The program will continue to execute under the supervision
-of Sandboxie, even after Administrator privileges have been granted.
+of Sandboxie, even after Administrator privileges have been granted.But if SysCallLockDown is not enabled, it could lead to a program escape from the sandbox!!!
```
offhub, line 1087:
```suggestion
of Sandboxie, even after Administrator privileges are granted. However,
if "SysCallLockDown" is not enabled, the program could escape from the sandbox!
```

DavidXanatos:
The underlying issue is that with admin privileges when talking to some windows components an adversarial application has more opportunity to exploit unfixed windows vulnerabilities.
While there is no report its almost certain that without RestrictDevices=y hackers may find exploits which allow kernel level access using some vulnerable not yet blacklisted driver.

The message as is is not quite right it should be more along the lines of the above explanation, may be:

Note:  The program will continue to execute under the supervision of Sandboxie, even after Administrator privileges have been granted. Please note that this may open more opportunities to the application to leverage a yet un-patched windows exploit or vulnerable driver and escape the sandbox. 

Also I was talking with yeyixiao and we came the the conclusion that the best option would be to drop the windows UAC prompt and instead present a sandboxie own dialog, offering a 3rd option that is to grant only faked admin rights to the process, and explain the security implications of each option.
I will be looking into this for one of the next builds.
Until then we can update the message along the lines of my above suggestion.

Also that said it should be noted that if there were to be a code execution vulnerability in one of the unprivileged ntoskrnl or win32k syscalls no mater the preset an application could leverage that to take over the system, but on a bright note that's something MSFT would patch very very very fast as that's a general thread for all of window's security model. So for Sandboxie to be reliable the users must keep their systems up to date.

Sandboxie/msgs/Sbie-English-1033.txt
```diff
@@ -1084,7 +1084,8 @@ User Account Control pop-up window.
 
 3243;txt;01
 Note:  The program will continue to execute under the supervision
-of Sandboxie, even after Administrator privileges have been granted.
+of Sandboxie, even after Administrator privileges are granted. However,
+if "SysCallLockDown" is not enabled, the program could escape from the sandbox!
```
offhub, line 1088:
```suggestion
of Sandboxie even after Administrator privileges are granted. However,
granting elevated access may increase the risk of the application
exploiting an unpatched Windows vulnerability or a vulnerable driver,
potentially allowing it to escape the sandbox.
```

offhub:
This warning is not very useful when UAC is set to the default level (secure desktop) or higher, because the system switches to the secure desktop before the message can be read. [^1]

[^1]: [Sandboxie & UAC](https://sandboxie-website-archive.github.io/www.sandboxie.com/old-forums/viewtopicb324b324.html?f=17&t=22250#p116676)

love-code-yeyixiao:
Then we could add a deliberate delay after start the elevated process

DavidXanatos:
Sandboxies own UAC is now implemented: https://github.com/sandboxie-plus/Sandboxie/commit/ec65bcbaf6529fbf3df4ec540232825949984940
With that I think wen can close this

PR #4927, opened 2025-07-27
Add option to use the new Qt Windows 11 style on SandMan

LumitoLuma:
Qt 6.7 introduced a new Windows 11 style which is, in my opinion, a bit broken and doesn't fit well with the rest of the Windows apps. It is the default style for apps running on Windows 11, and does not show on Windows 10.

Thus, I have disabled it by default and added an option to enable it, only on Windows 11.


DavidXanatos:
@LumitoLuma 
I found a solution to building a windows 7 driver with the new 2022 runner,
so we only are missing a fix for Qt 6 not running on 7, would you have time to set up a CI for Qt6 with these patches: https://github.com/crystalidea/qt6windows7
then we could bring back windows 7 compatibility.

PR #4956, opened 2025-08-12
Fix UAC prompt dialog in higher DPI

### LINKED ISSUE #4953: Bad UAC prompt layout at higher DPI in secure desktop

Make sure your display scaling factor is greater than 100%, for example, 125%. This results in a DPI greather than 96.

Launch something that requires Administrator's privilege.

In a secure desktop, it looks like this: (not working)

Image

When secure desktop is disabled, it looks like this: (working, but blurry)

Image

This was claimed to be fixed in #4863, but when secure desktop is enabled, it's still broken.

_No response_

### Expected behavior

The dialog should scale with the system DPI setting properly, without blurriness or broken layout.

### Affected program

Not relevant

### Download link

Not relevant

Not relevant to my request.

No, not at all.

### Crash dump

_No response_

Sandboxie-Plus v1.16.2

_No response_

Not relevant to my request.

Windows 10 Pro 22H2 64-bit (virtual machine)

Not relevant to my request.

None

_No response_

### Trace log

_No response_

```ini

```

gexgd0419:
Fixes #4953.

Changes:

- The call to `InitDpiAwareness` and its definition are removed. Setting DPI awareness to "unaware" is no longer needed.
- Every hard-coded value in pixels is replaced with `MulDiv(value, dpi, 96)` to take DPI into account.
- `hFontButtonText` is added in `SDialogParams` to set the text font for the buttons.
- The parent window of the UAC prompt will be disabled temporarily to make it more like a model dialog. This also prevents users from clicking the background and losing the focus on the dialog.

We can also consider using a different font than hard-coded "Arial", which might not play well with every language.

### DIFF
--- a/Sandboxie/apps/start/start.cpp
+++ b/Sandboxie/apps/start/start.cpp
@@ -369,40 +369,6 @@ _FX WCHAR *Get_Default_Browser(void)
 
 
 //---------------------------------------------------------------------------
-// InitDpiAwareness
-//---------------------------------------------------------------------------
-
-typedef DPI_AWARENESS_CONTEXT (WINAPI *PFN_SetThreadDpiAwarenessContext)(DPI_AWARENESS_CONTEXT);
-
-static BOOL InitDpiAwareness()
-{
-    HMODULE hUser32 = GetModuleHandleW(L"user32.dll");
-    if (!hUser32) {
-        // Should never happen
-        return FALSE;
-    }
-
-    PFN_SetThreadDpiAwarenessContext pSetThreadDpiAwarenessContext =
-        (PFN_SetThreadDpiAwarenessContext)GetProcAddress(hUser32,
-            "SetThreadDpiAwarenessContext");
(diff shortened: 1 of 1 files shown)


gexgd0419:
Here are the UAC prompt appearance on my system after the fix.

## 100%


## 125%


## 150%


## 175%

gexgd0419:
So this is the Simplified Chinese version of this dialog:

image

It's not supported by Arial, so it falls back to SimSun (宋体). However, it's no longer the default UI font, so it looks kind of weird.

Can we use the system default font instead, so the text seems more similar to the UI text in other places?

I'd suggest writing the message text in a single line, and let the text drawing code handles line break. Otherwise the difference in font character widths can mess up the manual line breaking, such as leaving a blank wrapped line in the 150% image above.

gexgd0419:
@DavidXanatos Do you have any thoughts about whether we should use the system default UI font instead and remove manual line breaks in the message text?

Those aren't implemented here, but if you think they're fine, I'll try and make a PR.

PR #4965, opened 2025-08-15
Translated some strings for ukrainian

Nazar1ky:


Nazar1ky:
Thank you for merging it!

Can I ask, if I can get Developer License?

Issue #1311, opened 2021-10-19
[0.9.8b] Plus is crashing windows 11 while trying to access Chrome


isaak654:
### Originally posted by @BilleBarrett in https://github.com/sandboxie-plus/Sandboxie/discussions/1308#discussioncomment-1504721
This update Plus is crashing windows 11 whI tried to access Chrome and Ms Chrome
The Classic version will not allow Chrome to open up. i.e. 94 & 95
How to revert back to previous version

BilleBarrett:
BTW on the new system I just build the newer version run with not change from the install.

 

Barrett Computers [Remote Desktop Services]
517 Miltwood Dr,  
Greensboro, NC 27455-2709
(336) 988-0891
"Your operating system and Data are only as good as your last backup"
Life is not a journey to the grave with the intention of arriving safely in a pretty and well-preserved body, but rather to skid in broadside, thoroughly used up, totally worn out,
and loudly proclaiming... 
 'Wow! What a ride!’
 <https://www.facebook.com/DesktopRemoteServices/> https://www.facebook.com/DesktopRemoteServices/

DavidXanatos:
Ok, i have revieved the driver changes and I dont see how thay cause a BSOD, so the next thing to test is on the affected system if you change back to the new driver is the crash reproducable?

DavidXanatos:
and if you install the previouse build it works fine,
did you kept the sandboxie.ini?
tra saving it to a backup location and running the new builf with an empty file.
if that works that its some option set there, in which case please send me the affected ini for inspection, if i can reproduce the issue here i can fix it.

BilleBarrett:
Right now I’m running Sandboxie 5.53.0, I’ll install Plus later to night 

 

Barrett Computers [Remote Desktop Services]
517 Miltwood Dr,  
Greensboro, NC 27455-2709
(336) 988-0891
"Your operating system and Data are only as good as your last backup"
Life is not a journey to the grave with the intention of arriving safely in a pretty and well-preserved body, but rather to skid in broadside, thoroughly used up, totally worn out,
and loudly proclaiming... 
 'Wow! What a ride!’
 <https://www.facebook.com/DesktopRemoteServices/> https://www.facebook.com/DesktopRemoteServices/

BilleBarrett:
Good morning David

I uninstalled everything and save the ini files 

1st I have installed 9.8.0 and made some setting  Sandboxie.9.8

I have just installed 9.8b and so far it is running

[...]

PR #5139, opened 2025-12-14
nitpick inconsistent whitespace

divinity76:
they all follow the same
^#\sdigit

pattern.. except 4 who didn't. 
This did actually cause some minor annoyance in a translation I'm working on.

>

### DIFF
--- a/Sandboxie/msgs/Sbie-English-1033.txt
+++ b/Sandboxie/msgs/Sbie-English-1033.txt
@@ -234,7 +234,7 @@ SBIE1309 Denied attempt to change account password
 # SBIE1310 Extended features are disabled until the license is reactivated
 # .
 
-#1311;pop;inf;01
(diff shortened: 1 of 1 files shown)


divinity76:
@offhub invalid?

divinity76:
@offhub `Visual Studio Code` is a very popular code editor from Microsoft,
and in VSCode if you write stuff like `#1311;pop;inf;01`  , it will trigger VSCode's CSS color picker:
image
And when the Color picker is triggered, you are just 2 missclicks away from
accidentally writing some random value to the line,
and these 4 lines triggers the VSCode color picker, which to translators can be confusing, and at worst, make them accidentally write some garbage values. It almost happened to me!


As for consistency, the file contains 150 lines matching `^#\s\d` but only 4 lines matching `^#\d` - the 4 i'm fixing in this PR.

PR #5141, opened 2025-12-14
Add Norwegian Bokmål translations

divinity76:
Add Norwegian Bokmål translations.
Translations were largely generated with the Argos Translate library https://github.com/argosopentech/argos-translate
using model https://huggingface.co/shethjenil/argostranslate/blob/main/translate-en_nb.argosmodel version 1.9

and corrected by a native Norwegian ( https://github.com/divinity76 / hans@loltek.net / Hans Henrik Bergan )

That saved a ton of time:
- Most of the translations were perfect
- Many required minor corrections
- Very rarely did I have to translate from scratch (iirc like 15 times in 4400 lines)

Still worked on this for 3 days though...

Among Norwegians who work in IT, there is a culture of mixing
English in Norwegian, but that was -not- done here.
For example, most people in IT would not write
"Feil:  %2"
"registeret",
"Windows-bygging" or "bygg",
they would just use the english words
"Error: %2"
"registry"
"Windows-build" or "build"

And Norwegian gamers don't use the term "spill", they use "Gaming".

but ofc, that is not valid Norwegian.

>


offhub:
Thank you for your contribution.

Please remove all TS files except `sandman_nb_NO.ts`.

Additionally, either use the original files as provided, or if you need to regenerate them use the following command. This command updates the existing `sandman_nb_NO.ts` file in place and does not create a new one:

```
lupdate SandMan.pri ../MiscHelpers/MiscHelpers.pri -ts sandman_nb_NO.ts
```
[sandman_nb_NO.zip](https://github.com/user-attachments/files/24152002/sandman_nb_NO.zip)

divinity76:
@offhub 
> Please remove all TS files except `sandman_nb_NO.ts`.

Done (also git was being moody about it, resorted to a force push)

> Additionally, either use the original files as provided, or if you need to regenerate them use the following command. 

I re-generated the sandman_nb_NO.ts with your command, and it did generate differently, thanks,
 but don't know what you mean by `use the original files as provided`, do you mean "leave the other ts files alone", or do you mean i touched something else I wasn't supposed to? there is no pre-existing sandman_nb_NO.ts , it's brand new in this PR.

offhub:
@divinity76 

Thanks again for your contribution to the project.

Regarding your question: When it's not necessary to change files for languages other than the one you are localizing, you don't need to worry about them, as the workflow automatically updates those files when needed.

divinity76:
```py
import re
from argostranslate import translate as argos_translate


class NbTranslator:
    _TOKEN_REGEX = re.compile(
        r"(%\d+|%[A-Za-z]|%\w+%|\{#[^}]+\}|\\n|\\t|%n|&lt;[^&]+?&gt;|</?[^>]+>|&[A-Za-z]+;)"
    )

    def __init__(self):
        languages = argos_translate.get_installed_languages()
        self._from_lang = next(lang for lang in languages if lang.code == "en")
        self._to_lang = next(lang for lang in languages if lang.code.startswith("nb"))
        self._translation = self._from_lang.get_translation(self._to_lang)

    def translate(self, text: str) -> str:
        if text is None:
            return text
        if not text.strip():
            return text

        segments = self._split_tokens(text)
        pieces = []
        for kind, value in segments:
            if kind == "token":
                pieces.append(value)
                continue

            if not value.strip():
                pieces.append(value)
                continue

            leading_len = len(value) - len(value.lstrip())
            trailing_len = len(value) - len(value.rstrip())

            core_start = leading_len
            core_end = len(value) - trailing_len if trailing_len else len(value)
            core = value[core_start:core_end]

            translated_core = self._translation.translate(core) if core else ""
            translated = value[:leading_len] + translated_core + value[core_end:]
            pieces.append(translated)
        return "".join(pieces)

    @staticmethod
    def _split_tokens(text):
        segments = []
        last = 0
        for match in NbTranslator._TOKEN_REGEX.finditer(text):
            start, end = match.span()
            if start > last:
                segments.append(("text", text[last:start]))
            segments.append(("token", match.group(0)))
            last = end
        if last < len(text):
            segments.append(("text", text[last:]))
        return segments


if __name__ == "__main__":
    import sys

    if len(sys.argv) != 3:
        print(f"Usage: {sys.argv[0]} <input> <output>")
        sys.exit(1)

    translator = NbTranslator()
    with open(sys.argv[1], encoding="utf-8", newline="") as fin, open(sys.argv[2], "w", encoding="utf-8", newline="") as fout:
        for line in fin:
            lineStripped = line.strip()
            doTranslation = len(lineStripped) > 1 and not lineStripped.startswith("#") and not lineStripped[0].isdigit()
            if doTranslation:
                line = translator.translate(line)
            fout.write(line)
```

PR #5157, opened 2025-12-29
update sandman_zh_CN.ts and sandman_zh_TW.ts

banlanzs:
>

Hello, I have a question. As shown in the image, I noticed that these translations already exist, but they are still marked as "unfinished."  Does this mean they haven't been used, or were they accidentally overlooked?
image


DavidXanatos:
this means the original english string has changed so the translation may be in need of updating

PR #5262, opened 2026-02-28
Add private network traffic blocking feature

love-code-yeyixiao:
Add a `BlockPrivateNet` boolean field to the `WFP_PROCESS` structure to enable per-process blocking of private network traffic.

Implement the `WFP_isPrivateNet` function to detect whether a destination IP belongs to a private or local network, supporting both IPv4 and IPv6 ranges: IPv4 private ranges (10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16, 169.254.0.0/16 via IPv4-mapped IPv6), and IPv6 local scopes (FC00::/7 for Unique Local, FE80::/10 for link-local).

Update the `WFP_UpdateProcess` function to read the `BlockPrivateNet` configuration key and store it in the process runtime state.

Modify `WFP_classify` to block traffic when the destination matches a private network range and `BlockPrivateNet` is enabled, independently of other blocking policies (`BlockInternet`, `BlockLoopback`), alongside existing custom firewall rule checks.


Kizuno18:
I reviewed the current draft head (`2e2dec5a149080b3a7c4f3b6c1ef5e8c62db9a35`) and found two concrete API-contract problems in the user-mode fallback.

First, `WSA_WSARecvFrom` applies the new private-address check immediately after the real call even when an overlapped receive returned `SOCKET_ERROR` / `WSA_IO_PENDING`. At that point `lpFrom` has not been populated, so the hook inspects incomplete caller storage and can replace the required pending result with a generic `SOCKET_ERROR`. The synchronous `recvfrom` path also runs the new check after a failed receive and returns a rejection without setting a Winsock error.

Second, the `ConnectEx` hook uses `SOCKET_ERROR` (`-1`) for the new rejection path. `ConnectEx` has a Boolean return contract, so `-1` is observed as success. The existing firewall and strict-bind early returns in the same hook have the same problem.

I prepared a focused fix on top of this exact head:

- commit: https://github.com/Kizuno18/sbxie/commit/62d3ad7e
- comparison: https://github.com/sandboxie-plus/Sandboxie/compare/BlockLocalNet...Kizuno18:sbxie:codex/help-pr-5262

The fix preserves overlapped receive results until completion, post-filters only successful synchronous receives, returns `FALSE` from all `ConnectEx` rejection paths, and consistently sets the Winsock error through the imported `WSASetLastError`. The WFP driver path remains the enforcement boundary for overlapped inbound traffic; when WFP is unavailable, the DLL fallback cannot safely post-filter an overlapped buffer before its completion callback.

Validation: the Win32 `SandboxDll.sln` build compiled the changed `net.c` cleanly. The build then stopped at the host's existing resource dependency because `afxres.h` is unavailable in this Build Tools installation.

PR #5341, opened 2026-04-13
refactor: integrate resource statistics into main view, removing ResourceView and adding summary info display

cuijianzhuang:
image


cuijianzhuang:
@DavidXanatos It has been refactored based on your suggestions.

PR #5361, opened 2026-05-02
Add breakout target-box and matching improvements

### LINKED ISSUE #1907: Allow BreakoutProcess and ForceProcess to support paths

https://github.com/sandboxie-plus/Sandboxie/releases/tag/1.0.8:

> added "BreakoutProcess=program.exe", with this option selected applications can be started unboxed from within a box https://github.com/sandboxie-plus/Sandboxie/issues/1500
>     -- Note: "BreakoutFolder=some\path" is also available

#1905 made it clear only `BreakoutFolder` supports paths.

Let `BreakoutProcess` use paths too like in `BreakoutProcess=some\path\program.exe`.
Let `ForceProcess` use paths too like in `ForceProcess=some\path\program.exe`.

_No response_

### LINKED ISSUE #2058: BreakoutFolder should take priority over ForceFolder

* When `ForceFolder` and `BreakoutFolder` are used at the same time, there are some scenarios where `BreakoutFolder` is ignored and a process still gets executed inside of the sandbox
* From what I can see, when you have Steam on your system (installed outside Sandboxie), and you force your game library to run inside of Sandboxie (in this case `D:\Games`), any Steam game (e.g. `D:\Games\...\SomeGame.exe`) will call Steam and Steam will launch `C:\Program Files (x86)\Common Files\Steam\steamservice.exe`, the later one installs missing dependencies (like DirectX), however `steamservice.exe` and all of it's child processes get executed inside Sandboxie despite `BreakoutFolder` being used.
* If I manually open Windows Explorer inside Sandboxie and manually launch Steam `C:\Program Files (x86)\Steam\steam.exe`, it launches outside Sandboxie as expected.
* If I remove `ForceFolder=D:\Games` and target the game I want to play instead `ForceProcess=SomeGame.exe`, the issue no longer happens (i.e. `steamservice.exe` executes outside of the sandbox as expected). Unfortunately this is not a good solution as it can get difficult to configure when you have lots of games.

Whenever a game has missing dependencies (i.e. DirectX, VC Redist, etc). Steam will try to install these dependencies through steamservice.exe (located under "C:\Program Files (x86)\Common Files\Steam") but steamservice.exe will open inside of the sandbox despite BreakoutFolder being used.

### Affected program

Steam

### Download link

Not relevant

The program is installed only outside the sandbox.

### Expected behavior

BreakoutFolder should take precedence when used alongside ForceFolder

Windows 10 21H2 x64

User account with secure desktop turned off for UAC prompts.

None, Windows Defender disabled

### What version of Sandboxie are you running?

Plus 1.1.3 x64

### Is it a regression?

_No response_

In an Application Compartment sandbox with no isolation (green sandbox icon).

### Can you reproduce this problem on an empty sandbox?

I can confirm it also on an empty sandbox.

_No response_

### Crash dump

_No response_

### Trace log

_No response_

```shell
#
# Sandboxie-Plus configuration file
#

[GlobalSettings]
FileRootPath=\??\%SystemDrive%\Sandbox\%USER%\%SANDBOX%
SeparateUserFolders=y
KeyRootPath=\REGISTRY\USER\Sandbox_%USER%_%SANDBOX%
IpcRootPath=\Sandbox\%USER%\%SANDBOX%\Session_%SESSION%
NetworkEnableWFP=n
EnableObjectFiltering=y
EnableWin32kHooks=y
EditAdminOnly=n
ForceDisableAdminOnly=n
ForgetPassword=n
Template=7zipShellEx
Template=WindowsRasMan
Template=WindowsLive
Template=OfficeLicensing
ForceDisableSeconds=60

[UserSettings_<REDACTED>]
SbieCtrl_AutoStartAgent=SandMan.exe
BoxDisplayOrder=Example
SbieCtrl_EnableAutoStart=y
SbieCtrl_UserName=user
SbieCtrl_NextUpdateCheck=1657211782
SbieCtrl_WindowCoords=200,150,1237,632
SbieCtrl_ActiveView=40021
SbieCtrl_ProcessViewColumnWidths=250,70,300

[Example]
Enabled=y
AutoRecover=y
BlockNetworkFiles=y
RecoverFolder=%{374DE290-123F-4565-9164-39C4925E467B}%
RecoverFolder=%Personal%
RecoverFolder=%Desktop%
BorderColor=#00ffff,ttl,6
Template=OpenBluetooth
Template=SkipHook
Template=FileCopy
Template=qWave
Template=BlockPorts
Template=LingerPrograms
Template=AutoRecoverIgnore
ConfigLevel=9

# Path where Steam games are installed
ForceFolder=D:\Games
# Files allowed to run outside the sandbox
OpenFilePath=C:\Program Files (x86)\Steam
BreakoutFolder=C:\Program Files (x86)\Steam
BreakoutFolder=C:\Program Files (x86)\Steam\*
OpenFilePath=C:\Program Files (x86)\Common Files\Steam
BreakoutFolder=C:\Program Files (x86)\Common Files\Steam
BreakoutFolder=C:\Program Files (x86)\Common Files\Steam\*

OpenIpcPath=$:steam.exe
OpenIpcPath=*\BaseNamedObjects\Steam*
OpenWinClass=SteamWinsockInitFakeClass_*

NoSecurityIsolation=y
```

### LINKED ISSUE #2121: Manual Sandbox Target Designation for BreakoutProcess/BreakoutFolder

It would be great If we can designate a manual sandbox target for `BreakoutProcess`/`BreakoutFolder`.

There are two use cases covered here:

`BreakoutProcess=test.exe` 
1.  `test.exe` will run unsandboxed if no force is set
2.  `test.exe` will run sandboxed if force is set

But If I don't want to use forced sandbox for breakout there is no way to set it.

`BreakoutProcess=test.exe`
- `test.exe` will run unsandboxed

`BreakoutProcess=test.exe,TESTBOX`
- `test.exe` will run in TESTBOX

#

`ForceProcess=test.exe`
`BreakoutProcess=test.exe`
- `test.exe` will run in the forced sandbox

`ForceProcess=test.exe`
`BreakoutProcess=test.exe,TESTBOX`
- `test.exe` will run in TESTBOX

_No response_

### LINKED ISSUE #5350: Per-box BreakoutProcess should always override Force rules

When specified within the same sandbox config (i.e. not globally), `BreakoutProcess` rules should always take precedence over `ForceFolder` and `ForceChildren` rules. Currently, the opposite implementation is true. 

To replicate:
1. Create a new sandbox with the rules `ForceChildren=explorer.exe` and `BreakoutProcess=cmd.exe`
2. Open File Explorer to "C:\Windows\System32" and launch "cmd.exe"
3. Observe that cmd launches sandboxed.

Always.

### Expected behavior

A process specified by `BreakoutProcess` in a sandbox config should always break out of that sandbox.

### Affected program

Not relevant

### Download link

Not relevant

Not relevant to my request.

No, not at all.

### Crash dump

_No response_

Sandboxie-Plus 1.17.4 64-bit

_No response_

In a standard isolation sandbox (yellow sandbox icon).

Windows 11 Enterprise 24H2

Not relevant to my request.

N/A

_No response_

### Trace log

_No response_

```ini

```

offhub:
Resolves #1907
Resolves #2058
Resolves #2121
Resolves #5350

May resolve some other [Breakout](https://github.com/sandboxie-plus/Sandboxie/issues?q=is%3Aissue%20state%3Aopen%20Breakout)-related issues as well.

---

#### Baseline and Compatibility

- 1.17.6 baseline behavior does not honor Priority, Recursive, or TargetBox metadata for `ForceProcess`, `ForceFolder`, `ForceChildren`, `BreakoutProcess`, `BreakoutFolder`, and `BreakoutDocument`.
- In 1.17.x, extension behavior is gated by `UseForceBreakoutRuleExtensions`.
- With `UseForceBreakoutRuleExtensions=n` (or unset), extension-capable settings keep 1.17.6-compatible parsing behavior (text after first | ignored).

#### Newly Added Settings

- `UseForceBreakoutRuleExtensions=[y|n]` (default disabled): enables or disables rule metadata parsing for `ForceProcess`, `ForceFolder`, `ForceChildren`, `BreakoutProcess`, `BreakoutFolder`, and `BreakoutDocument`.
- `BreakoutUseTargetDir=[process,][y|n]`: prefers the launched target directory as breakout working directory (global or per-process form).

#### Summary

- `BreakoutProcess`, `BreakoutFolder`, and `BreakoutDocument` support optional `TargetBox` metadata for explicit breakout routing.
- Priority-based arbitration is supported via `Priority=value` (lower number wins, higher precedence).
- Rule-extension parsing is gate-controlled: when `UseForceBreakoutRuleExtensions` is disabled or unset, metadata after the first `|` is ignored (`Priority`, `Recursive`, `TargetBox` are not applied).
- `Recursive` depth metadata is supported for folder-style force and breakout rules where applicable.
- `BreakoutProcess` supports image-name and path-pattern matching and still requires host-path validation before breakout is allowed.
- `BreakoutFolder` matching includes trailing-backslash normalization and improved wildcard behavior.
- Process breakout candidate detection includes targeted breakout entries; final force-vs-breakout arbitration and target routing are resolved service-side.
- `BreakoutDocument` is priority-aware and can arbitrate against `ForceProcess` and `ForceFolder`, and also against `BreakoutProcess` on the breakout side for target selection.
- `TargetBox` is resolved only from the winning breakout side, not from weaker matching breakout rules.
- Invalid explicit targets (missing/disabled, or same as source box) are rejected: process starts fall back to normal boxed creation, and document opens fall back to source-box handling.
- Source-box force capture can be skipped when breakout wins by priority, while force rules from other boxes can still win when their precedence is stronger.
- `BreakoutUseTargetDir` is honored in breakout launch paths to reduce re-force side effects caused by inherited sandbox working directories.

#### Explicit Precedence Warning

- Lower numeric `Priority` means stronger precedence.
- When extensions are enabled, a matching breakout rule with a lower numeric `Priority` can overrule matching `ForceProcess`, `ForceFolder`, and `ForceChildren` rules for that launch path.
- If no matched rule has explicit `Priority`, legacy defaults still apply by context:
1. Process-start path: force side wins by default.
2. Document-open path: breakout side wins by default.

#### `ForceChildren`-related behavior

- `ForceChildren` match origin and priority are tracked and exposed via `SBIE_FLAG_FORCED_CHILD_PROCESS`.
- `ForceChildren` lineage is inherited by descendants.
- `ForceChildren` lineage is preserved across model-pid starts, including `PCA` restart flow.
- Service-side breakout arbitration enforces `ForceChildren` lineage for both direct forced-child callers and sandboxed ancestor-chain matches.
- For `ForceChildren`-origin callers and lineage, breakout is denied unless breakout priority explicitly beats force-side priority.
- If breakout does not beat force-side priority, normal boxed behavior is retained.

### Implementation Scope

- Shared rule parsing, normalization, and policy-decision helpers were centralized for consistent DLL/SVC/DRV behavior.
- DLL-side breakout candidate and document pre-check paths were updated to hand off arbitration cleanly and preserve fallback behavior.
- Driver-side force matching gained extension-aware priority handling, breakout target ranking, recursive-depth handling, and `BreakoutDocument`-aware arbitration hooks.
- Service-side process and document arbitration now performs priority-aware winner resolution, winning-side target selection, and explicit-target validation/fallback handling.


DavidXanatos:
I am not sure if ```|TargetBox``` is the way to go, the way sbie.ini config lines are formatted is very basic, we should before adding more special cases may be think of a universal format change.
See case where that was needed ´´´RpcPortBinding=kernel32.dll,'0497b57d-2e66-424f-a0c6-157cd5d41700@ncalrpc:',Resolve=AppInfo,TimeOut=y´´´
I am thinking if we should not think of a universal approach, I think being able to append additional parameters like an explicit rule priority or what not would be great.
That said using ```|``` is not bad, its a forbidden symbol for paths and in sbie.ini it is almost unused, except before a path to indicate to not append a default ```*``` to the path end. So everyone else its free to be used.

I mean at this opportunity we should think of how we want to proceed such that we don't end up with a dozen different conventions depending on for which setting they are.
At least we should use ```|TargetBox=TargetBoxName``` so that we can add more options separated with ```|``` for example we may want to configure on force folder if that directive should include sub folders or now, etc...

What do you think?

offhub:
I agree that we should standardize on an extension format at least for the Force* and Breakout* settings and keep it consistent there. That would already make it easier to support additional parameters without introducing per-setting syntax variations.

So far I avoided `key=value` mainly because it was quicker to write while testing, but for a proper design it makes more sense to adopt a structured approach.

With that in place, we could unify handling of Force* and Breakout* under a single parsing/processing path, and control behavior through extensions instead. For example:

    |TargetBox=TargetBoxName
    |Recursive=y
    |Priority=100

This would let us handle things like precedence and recursion in a consistent way without adding new settings or special-case parsing.

Kizuno18:
I found a UI/model mismatch in the breakout rule editor. `OptionsForce.cpp` reads and writes Priority, Recursive, and Target Box through columns 2–4, but `treeBreakout` in `OptionsWindow.ui` declares only Type and Name.

I verified the exact behavior with Qt 6.8.3 in an offscreen `QTreeWidget` probe: calling `item->setText(4, ...)` leaves both `tree.columnCount()` and `tree.model()->columnCount()` at 2, and `model()->index(0, 4)` is invalid. The extra values can live on the item, but the user cannot see or edit those columns. This blocks configuring the new breakout extensions through the table, especially `TargetBox`.

The narrow fix adds the three missing breakout headers to match the existing column constants and the force tree. The updated UI parses successfully with Qt `uic`, which emits all five `treeBreakout` headers:

https://github.com/Kizuno18/sbxie/commit/aece92bf64faeb2cebff228035dd0c4318b60608

PR #5373, opened 2026-05-08
Change GetArguments parameter from wstring to wstring_view

fabiodefilipposoftware:
Replace wstring& with wstring_view to improve security

>


DavidXanatos:
this does not compile!
also using wstring_view has no security impact it only improves performance by avoiding full string copying.

PR #5374, opened 2026-05-08
Resolve potentially loosing pointer with realloc

fabiodefilipposoftware:
If realloc fails due to lack of memory, the original pointer is lost (causing a memory leak) and the application crashes on the next write attempt.

>


DavidXanatos:
wtf? newAllocatedLength should be allocatedLength etc... this also doe snot compile as is.

PR #5409, opened 2026-05-29
Remove unsafe inline WFP trace logging block

woolkingx:
## Summary

This PR removes an inactive `NetFwTrace` logging block from `Sandboxie/core/drv/wfp.c` inside the WFP classify callback.

The removed block formatted monitor strings and called `Session_MonitorPut` from the classify path. The existing comment already noted that this path can run at `DISPATCH_LEVEL`, while the monitor path uses pageable memory. Keeping the disabled implementation in place made it easy to re-enable an IRQL-unsafe logging path later.

This PR changes that path to preserve the current traffic decision behavior while making the boundary explicit:

- remove the unused `log` local and assignment
- remove the commented inline `RtlStringCbPrintfA` / `Session_MonitorPut` implementation
- leave a compact contract comment requiring nonpaged classify-side capture plus a `PASSIVE_LEVEL` monitor writer before traffic logging is enabled from this callback

It does not implement WFP traffic logging.

## Rationale

A WFP classify callback is a decision path, not a safe place to perform pageable monitor logging.

If the old commented implementation were re-enabled, traffic tracing could become an IRQL/DDI bugcheck path because it combines classify-time execution with string formatting and session monitor writeback. Removing the inactive code keeps the current no-logging behavior, but prevents a stale workaround from looking like the intended implementation.

The correct future shape is a deferred logger: capture only resident, bounded data from the classify path, then format and write monitor records from an executor that is legal for the monitor subsystem.

## Official/API shape

Microsoft documents WFP classify callbacks as running at `IRQL <= DISPATCH_LEVEL`:

- https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/fwpsk/nc-fwpsk-fwps_callout_classify_fn0

Microsoft documents `RtlStringCbPrintfA/W` as requiring resident strings when used above `PASSIVE_LEVEL`:

- https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntstrsafe/nf-ntstrsafe-rtlstringcbprintfa

The monitor path reached by `Session_MonitorPut` takes the session resource path and writes to the monitor log. That is not the same IRQL contract as the WFP classify callback.

## Verification

- `WFP_classify` no longer contains direct `Session_MonitorPut` / `Session_MonitorPutEx` calls
- `WFP_classify` no longer contains direct `RtlStringCbPrintfA` / `RtlStringCbPrintfW` calls
- the unused `log` local and `wfp_proc->LogTraffic` assignment were removed from this path
- traffic block/permit decision code is otherwise unchanged
- `git diff --check` passes

- driver build
- Driver Verifier with IRQL checking / DDI compliance enabled
- IPv4/IPv6 TCP and UDP traffic, loopback and non-loopback traffic, blocked and permitted paths
- `NetFwTrace` enabled and disabled compatibility observation
- future logger implementation should prove no inline monitor write, no verifier violation, no monitor-log corruption, and no changed block/permit decisions

### DIFF
--- a/Sandboxie/core/drv/wfp.c
+++ b/Sandboxie/core/drv/wfp.c
@@ -876,7 +876,6 @@ void WFP_classify(
 		UINT16 remote_port = inFixedValues->incomingValue[remotePortIndex].value.uint16;
 
 
-		BOOLEAN log = FALSE;
 		BOOLEAN block = FALSE;
 		BOOLEAN noloop = FALSE;
 		BOOLEAN isloopback = FALSE;
@@ -895,7 +894,6 @@ void WFP_classify(
 		wfp_proc = map_get(&WFP_Processes, processId);
 		if (wfp_proc) {
 
-			log = wfp_proc->LogTraffic;
 			block = wfp_proc->BlockInternet;
 			noloop = wfp_proc->BlockLoopback;
 			isloopback = WFP_isLoopback(&remote_ip);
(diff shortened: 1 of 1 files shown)


woolkingx:
Closing this PR because it only removes the stale unsafe inline WFP trace logging block. That is a cleanup/precondition, not a meaningful SREV-027 fix by itself. The useful fix should provide an executable deferred WFP trace path: nonpaged classify-side capture, a bounded DISPATCH-safe queue/ring, and a PASSIVE_LEVEL writer to Session_MonitorPutEx with Windows Driver Verifier runtime proof.

PR #5410, opened 2026-05-29
Defer WFP NetFwTrace logging out of classify callback

woolkingx:
## Summary

This PR defers WFP `NetFwTrace` monitor logging out of `Sandboxie/core/drv/wfp.c`'s `WFP_classify` callback.

The old in-source `NetFwTrace` block was commented out, but it showed the unsafe implementation shape: format the traffic string and call the session monitor path inline from the WFP classify callback.

This PR changes that path to preserve the WFP classify boundary:

- `WFP_classify` captures only fixed traffic facts after reading the per-process `LogTraffic` flag
- trace records are copied into a bounded nonpaged `WFP_TRACE_RECORD` queue
- a system thread drains the queue and formats the monitor string outside classify
- `Session_MonitorPut` is called only from the deferred writer path
- WFP load/unload starts, stops, drains, and frees the trace worker/queue

It does not change the traffic block/permit decision path.

## Rationale

A WFP classify callback can run at `DISPATCH_LEVEL`. The monitor logging path is not a safe boundary to call inline from that callback, and the previous commented block also performed string formatting in the same path.

Leaving the old block as the apparent implementation route makes it easy to re-enable `NetFwTrace` later with an IRQL-sensitive edge:

```text
WFP_classify
  -> RtlStringCbPrintfW
  -> Session_MonitorPut
  -> session monitor state
```

The safer shape is to keep classify-side work small and resident, then perform formatting and monitor writeback from a later executor:

```text
WFP_classify
  -> WFP_TraceEnqueueFromClassify
  -> bounded nonpaged queue
  -> WFP_TraceThreadProc
  -> WFP_TraceWrite
  -> Session_MonitorPut
```

## Official/API shape

Microsoft documents WFP `classifyFn` callbacks as callable at `IRQL <= DISPATCH_LEVEL`:

- https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/fwpsk/nc-fwpsk-fwps_callout_classify_fn0

Microsoft documents `RtlStringCbPrintf*` as callable at any IRQL only if strings being manipulated are always resident in memory; otherwise the caller must run at `PASSIVE_LEVEL`:

- https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntstrsafe/nf-ntstrsafe-rtlstringcbprintfa

The deferred path uses standard kernel event/thread primitives rather than doing monitor work in classify:

- https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-pscreatesystemthread
- https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-keinitializeevent
- https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-kesetevent
- https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-kewaitforsingleobject

## Verification

- `WFP_classify` no longer contains inline `Session_MonitorPut*` calls
- `WFP_classify` no longer contains inline `RtlStringCbPrintf*` calls
- `WFP_classify` enqueues through `WFP_TraceEnqueueFromClassify` after releasing `WFP_MapLock`
- the deferred path owns `WFP_TRACE_RECORD`, a bounded nonpaged queue, drop counter, event, worker thread, writer, start, and stop logic
- only `Sandboxie/core/drv/wfp.c` is changed in this PR branch
- `git diff --check upstream/master..HEAD` passes

- driver build
- Driver Verifier with IRQL checking / DDI compliance enabled
- `NetFwTrace` enabled and disabled
- IPv4/IPv6 and TCP/UDP traffic
- blocked/permitted and loopback/non-loopback traffic
- `BlockInternet`, `BlockLoopback`, and rule-based block/permit cases
- process churn and rule refresh while traffic is active
- monitor log readback for process id, address, port, protocol, direction, and action
- high-rate traffic to observe queue overflow/drop behavior
- WFP unload / driver unload drain behavior

### DIFF
--- a/Sandboxie/core/drv/wfp.c
+++ b/Sandboxie/core/drv/wfp.c
@@ -86,6 +86,8 @@ DEFINE_GUID(WPF_RECV_CALLOUT_GUID_V6,	// 0bf56435-71e4-4de7-bd0b-1af0b4cbb9f9
 #define WFP_FILTER_NAME L"SbieFilter"
 #define WFP_FILTER_DESCRIPTION L"A filter that uses by sandboxie to implement internet restrictions"
 
+#define WFP_TRACE_QUEUE_CAPACITY 256
+
 
 //---------------------------------------------------------------------------
 // Structures and Types
@@ -101,6 +103,18 @@ typedef struct _WFP_PROCESS {
 	LIST NetFwRules;
 
 } WFP_PROCESS;
+
+typedef struct _WFP_TRACE_RECORD {
+
+	HANDLE ProcessId;
+	BOOLEAN Send;
+	BOOLEAN V6;
+	BOOLEAN Block;
+	UCHAR Protocol;
+	UINT16 RemotePort;
+	IP_ADDRESS RemoteIp;
+
+} WFP_TRACE_RECORD;
 
 
 //---------------------------------------------------------------------------
@@ -120,6 +134,23 @@ const WCHAR* Process_MatchImageAndGetValue(BOX* box, const WCHAR* value, const W
 ULONG Process_GetTraceFlag(PROCESS *proc, const WCHAR *setting);
 
 void WFP_FreeRules(LIST* NetFwRules);
+
+BOOLEAN WFP_TraceStart(void);
+
+void WFP_TraceStop(void);
+
+void WFP_TraceEnqueueFromClassify(
+	HANDLE processId,
+	BOOLEAN send,
+	BOOLEAN v6,
+	const IP_ADDRESS* remote_ip,
+	UINT16 remote_port,
+	UCHAR protocol,
+	BOOLEAN block);
+
+void WFP_TraceThreadProc(PVOID StartContext);
+
+void WFP_TraceWrite(const WFP_TRACE_RECORD* record);
 
 #ifdef ALLOC_PRAGMA
 #pragma alloc_text (INIT, WFP_Init)
@@ -201,6 +232,17 @@ static BOOLEAN WPF_MapInitialized = FALSE;
 static map_base_t WFP_Processes;
 static KSPIN_LOCK WFP_MapLock;
 
+static WFP_TRACE_RECORD* WFP_TraceQueue = NULL;
+static ULONG WFP_TraceQueueRead = 0;
+static ULONG WFP_TraceQueueWrite = 0;
+static ULONG WFP_TraceQueueCount = 0;
+static ULONG WFP_TraceDropCount = 0;
+static KSPIN_LOCK WFP_TraceLock;
+static KEVENT WFP_TraceEvent;
+static HANDLE WFP_TraceThreadHandle = NULL;
+static BOOLEAN WFP_TraceStopping = FALSE;
+static BOOLEAN WFP_TraceReady = FALSE;
+
 
 //---------------------------------------------------------------------------
 // WFP_Alloc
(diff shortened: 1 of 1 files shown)


DavidXanatos:
That is very nice but needs to be tested very thoroughly, so expect it to be only merged in the over next build or so...

woolkingx:
This issue/PR is connected to several of my SREV items, not only the immediate WFP patch.

The current code patch mainly covers:

- SREV-027: WFP NetFwTrace must not write monitor logs inline from WFP_classify. The fix is to capture only fixed data in classify and defer formatting / Session_MonitorPut to a worker path.
- SREV-344: WFP transaction abort failure should be captured/logged instead of silently ignored during install cleanup.
- SREV-345: WFP rule-load failure logging ownership should be clear. WFP_LoadRules owns the allocation-failure log, while WFP_UpdateProcess owns fail-closed state and cleanup.

It is also related to these proof / boundary items:

- SREV-028: monitor ring readback boundary.
- SREV-220: MonitorGet2 bulk readback buffer floor.
- SREV-232: log_buffer allocation / size contract.
- SREV-239: WFP module lifecycle topology.
- SREV-319: socket creation vs WFP enforcement boundary.
- SREV-022: shared kernel runtime-capture proof model.

So before I continue splitting or expanding the work, I would like to ask how you prefer this to be handled.

Options:

1. I submit each SREV as a separate focused PR.
2. I submit one larger PR that contains the related WFP / monitor proof work together.
3. I submit a documentation PR under docs/ first, containing the SREV analysis, proof boundaries, and runtime test matrix, then follow with smaller code PRs.

My recommendation is option 3: add the SREV material as a docs PR first, so the review scope, runtime gates, and related boundaries are visible before we decide how much code should be merged in one patch.

DavidXanatos:
as long as all the changes touch the same mechanism/component one large PR if well tested is better

woolkingx:
I have completed a source-level core audit pass and collected the result in the audit ledger/report.

The current coverage table has 287 Sandboxie/core files, 276 reviewable core files, and 352 SREV ledger entries. All 276 reviewable core files are named in the ledger, with no unnamed reviewable core files left in the coverage table. There are also three local validation checkpoints.

The main focus of this work is Windows API compatibility and correctness: buffer sizes, counted strings, handle ownership, object lifetime, access rights, RPC/IPC wire shape, service/driver boundaries, user/kernel crossings, and similar correctness edges.

My original goal was to understand why Sandboxie can enter a system-wide hang without a BSOD. The audit did identify and classify many source-level correctness risks, but turning that into a real patch plan is harder. The involved paths are complex, and I do not yet have enough runtime confidence to claim a clean root cause or a complete fix.

The current material is therefore analysis/source-gate material, not a concrete patch plan. The table still shows 231 source-level items that need Windows runtime proof. I also tried setting up a test environment, but the result is not good enough yet; this needs a more complete Windows testing workflow and longer focused investigation.

If useful, I can package the full audit ledger, report, local validation checkpoints, and related prototype/code material for reference.

woolkingx:
I collected this PR, the SetupAPI PR, and the related SREV/audit material into a visible reference directory on the fork repo default branch:

https://github.com/woolkingx/Sandboxie

Entry point:

https://github.com/woolkingx/Sandboxie/tree/master/srev-reference-package

For this PR, the relevant material is SREV-027 / SREV-344 / SREV-345 under `srev-reference-package/audit/ledger/`, plus the patch file under `srev-reference-package/patches/`. This is source/reference material, not a claim that Windows runtime proof is complete.

You can cherry-pick, split, rewrite, or close this PR as you prefer. I will stop keeping this as an active upstream-decision blocker on my side.

woolkingx:
Closing this PR from my side. Details and reference material are in the fork repo default branch:

https://github.com/woolkingx/Sandboxie

Reference package directory:
https://github.com/woolkingx/Sandboxie/tree/master/srev-reference-package

woolkingx:
Update: I rewrote the fork default branch into a single public reference commit on top of the original fork baseline, so the complete code delta can be reviewed directly from one commit diff.

Fork commit:
https://github.com/woolkingx/Sandboxie/commit/ca0d646a513e67acc3667fa318bbd4f71beddfdc

Reference package:
https://github.com/woolkingx/Sandboxie/tree/master/srev-reference-package

For this PR, the relevant material remains SREV-027, SREV-344, and SREV-345, plus the broader core API / Windows runtime-boundary review notes. The two patch exports are only focused PR references; the single fork commit is the complete archived source/reference snapshot. The fork is archived and is intended as reference material only.

PR #5431, opened 2026-06-17
Add Wildcard Recovery Patterns and Batch Message/Notification UI Updates

### LINKED ISSUE #3761: Wildcard Support for AutoRecoverIgnore

**Description**

I propose adding wildcard support to Sandboxie's `AutoRecoverIgnore` feature. Currently, it only allows exact path specifications, limiting flexibility in ignoring multiple files or directories based on patterns.

**Details**
- **Current Behavior:** `AutoRecoverIgnore` accepts only exact paths.
- **Proposed Enhancement:** Introduce wildcard expressions support to enable pattern-based ignoring of files or directories.

Users could specify `%Tmpl.Brave%\*Cache*` to ignore all cache-related files within `%Tmpl.Brave%`:

`AutoRecoverIgnore=%Tmpl.Brave%\*Cache*`

_No response_

### LINKED ISSUE #5278: Feature: Execlude files or folders from quick recovery

When a program not only writes requested user data but also it's own program data, the recovery window gets messy really quickly.

Like the execlude feature for immediate recovery.
I want to be able to set exclusion rules for file types and folders for quick recovery.

I don't know if an action (e.g. hide from the window, delete, keep) for each entry would be useful.

Instead of adding another setting, the already existing entries for immediates recovery could also be used for hiding these files in the quick recovery window.
They could be shown when 'show all files' is ticked.

### LINKED ISSUE #5318: Allow ignoring wildcard in file recovery

For example, `__pycache__` directory.

Think `.gitignore` syntax

Ignoring `*.pyc`

offhub:
Resolves #3761
Resolves #5278
Resolves #5318

Introduce wildcard matching for RecoverFolder and AutoRecoverIgnore patterns across DLL and UI layers.

- Add common/recovery_wildcard.h: shared wildcard engine (supports *, ?, **) with path-separator-aware semantics.
- Implement wildcard matching and network/NT/UNC alias handling in core DLL (core/dll/file_recovery.c); fall back to fast literal checks when no wildcards. Handles relative vs full-path roots and avoids overflow on large paths.
- Update SandMan to use real/normalized paths for quick/instant recovery, filter immediate/pop-up recover notifications via CRecoveryWindow::IsFileIgnored, and keep quick recovery UI counts for filtered vs unfiltered files.
- UI changes: OptionsWindow adds UseAutoRecoverIgnoreForQuick checkbox (default on); RecoveryWindow adds Show Ignored Files checkbox and integrates wildcard recover-folder handling (wildcards are not scan roots). PopUpWindow now passes normalized BoxPath to recovery entries.
- Update SbieSettings.ini and CHANGELOG.md documenting wildcard support, new option, and related fixes.

This change improves matching accuracy for device/redirector paths (NT vs UNC) and lets administrators express flexible recovery/exclude rules.


DavidXanatos:
Why use a new wildcard matching code? 
we have Sandboxie\common\pattern.c like this
```
    PATTERN *pattern = Pattern_Create(NULL, pat, TRUE, 0);
    BOOLEAN match = Pattern_Match(pattern, testlwr, wcslen(testlwr));
    Pattern_Free(pattern);
```
the pattern itself can be created on start and kept in RAM for better performance.

I think its more elegant to use the existing mechanisms instead of duplicating functionality.

offhub:
Are you referring to the DLL only, or both the DLL and SandMan?

One reason I used a separate matching is that I was thinking about possibly adding additional matcher types in the future, such as regex-style matching.

That said, I could switch the DLL (or both) side to use the common pattern instead.

DavidXanatos:
I mean the DLL, in sandman you can use QRegularExpression::wildcardToRegularExpression.

regex-style matching Is something that if added should be added to all pattern matching mechanisms but the difficulty is how do we tell a regexp pattern from a normal pattern apart. We would need some prefix or dedicated syntax that would be unambiguous and not collide with legit wildcard path patterns.

DavidXanatos:
PS: thinking about it ```<``` and ``` > ``` would be a good marker so ```FileOpenPath=C:\<Program Files(?: \(x86\))?>\Program\*```
So we would allow parts of the pattern to be a regexp. the current pattern mechanism already splits the path in constant sections separated by * this approach would introduce a new section type which would do a regexp matching, also we would need to indicate if between 2 sessions a wildcard is allowed or if they must be directly adjacent.

PR #5465, opened 2026-07-14
Update lang_de.json

APMichael:
- Few corrections.


APMichael:
@offhub I think the corrections to the English strings will also be useful for your pending pull request #5462.

PR #5488, opened 2026-07-26
Add portable sandbox support to NewBoxWizard

love-code-yeyixiao:
First draft by DeepSeek

In the `CNewBoxWizard::TryToCreateBox` method, add support for portable sandboxes, allowing users to select a directory to store configuration files, and write the file when creating the sandbox. Modified the sandbox creation logic to accommodate the requirements of portable sandboxes, and added status checks to ensure the creation process is error-free.

In the `CBoxTypePage` constructor, add a checkbox for users to select creating a portable sandbox, and provide related tooltips.

In the `CSummaryPage::initializePage` method, add a description for portable sandboxes, informing users that their configuration will be stored in an external .ini file.


Kizuno18:
The portable creation path can destroy an existing config before it knows whether the new box is usable. `QFile::WriteOnly` truncates an existing `<BoxName>.ini`, and both the `ImportBox` append failure path and the later rollback then remove that same file.

I prepared a data-preserving fix in [863316dc](https://github.com/Kizuno18/sbxie/commit/863316dc7cab07e2fdafed969f5a64a371432325). It creates the ini exclusively, checks directory creation and the complete write, and keeps the file when rollback of the global `ImportBox` setting cannot be confirmed. It also propagates the actual append/reload/rollback status instead of discarding it through `UpdateTextList`, which always returns success.

I verified the location behavior against `Conf_Import_Include`: the driver injects `FileRootPath` from the imported ini path with `.ini` removed, so a selected `Portable\\Portable.ini` stores data under `Portable\\Portable` as intended.

Validation used Qt 6.8.3 and MSVC 17.14. `NewBoxWizard.cpp` compiles successfully through its generated release makefile. A focused test built against the production `CreatePortableBoxIni` helper passes three cases in a `QTemporaryDir`: an existing ini remains byte-for-byte unchanged, a new ini has the expected section, and creation fails without leaving a file when the parent directory is missing.

PR #5489, opened 2026-07-26
Add command-line flag injection

### LINKED ISSUE #5487: Administative programs based on the tool "fltmc" cannot run on a yellow box

In fact, some programs uses the CLI `fltmc` to check if they have administrative prvllege.
No matter if you have grant it administrator privlledge, no matter whether you uses Drop/FakeAdminRight=n or not, it believes you that you're not administrative.

_No response_

### Expected behavior

When Drop/FakeAdminRight=n, it should be able to let the program believe that it has administrative privlleges.

### Affected program

Not relevant

### Download link

Not relevant

The program is installed only inside a sandbox (NOT in the real system anyway).

No, not at all.

### Crash dump

_No response_

1.18.0

I have been using the same version for some time.

I don't think so

In a standard isolation sandbox (yellow sandbox icon).

Win11 25H2

Not relevant to my request.

Defender

_No response_

### Trace log

_No response_

```ini

```

love-code-yeyixiao:
In `kernel.c`, the functionality to inject command-line flags has been implemented.
This functionality checks the command-line arguments of the current process and injects specified flags based on the `InjectCmdLine` settings in the configuration file. At the same time, it hooks the `GetCommandLineW` and `GetCommandLineA` functions to ensure that the new command line can be correctly retrieved.

### DIFF
--- a/Sandboxie/core/dll/kernel.c
+++ b/Sandboxie/core/dll/kernel.c
@@ -221,6 +221,62 @@ _FX BOOLEAN Kernel_Init()
 		}
 	}
 
+	//
+	// InjectCmdLine: inject command-line flags into any process
+	// Format: InjectCmdLine=process_name.exe,flags to inject
+	//
+	if (!Kernel_CommandLineW.Buffer) {
+		RTL_USER_PROCESS_PARAMETERS* ProcessParms = Proc_GetRtlUserProcessParameters();
+
+		if (!wcsstr(ProcessParms->CommandLine.Buffer, L" --type=")) {
+
+			WCHAR buf[CONF_LINE_LEN];
+			ULONG index = 0;
+			while (1) {
+				NTSTATUS status = SbieApi_QueryConfAsIs(NULL, L"InjectCmdLine", index, buf, ARRAYSIZE(buf));
+				if (!NT_SUCCESS(status)) break;
+				++index;
+
+				WCHAR* ptr = wcschr(buf, L',');
+				if (!ptr) continue;
+
+				*ptr++ = L'\0';
+
+				if (_wcsicmp(Dll_ImageName, buf) == 0) {
+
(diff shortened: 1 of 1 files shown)


DavidXanatos:
why not when --type= is in teh command line and calling SBIEDLL_HOOK multiple times on the saem function inside a loop i s a bug,
did this even work or crash?

love-code-yeyixiao:
a "break;" is in the end of the loop so it will just get one time to be executed.&nbsp;
The "if" was designed to skip arguments that might be operated by AddCustomChroiumFlag.

Kizuno18:
I confirmed two failure modes in the new injection block. `UNICODE_STRING::MaximumLength` is a 16-bit byte count, so adding a fixed `(CONF_LINE_LEN + 8) * sizeof(WCHAR)` reserve to the process parameter value can wrap. The resulting allocation is then used without checking `LocalAlloc`, and `GetCommandLineA` is hooked even if `RtlUnicodeStringToAnsiString` fails and leaves no ANSI buffer.

The outer `--type=` test also conflicts with a generic per-process rule: an explicitly configured non-Chromium process is skipped solely because one of its arguments contains that substring. The specialized `CustomChromiumFlags` path already takes precedence through `Kernel_CommandLineW.Buffer` when it applies.

I prepared a narrow fix in [8a3eff9cefcbe101eb43fd6c9f88a6dba80c1744](https://github.com/Kizuno18/sbxie/commit/8a3eff9cefcbe101eb43fd6c9f88a6dba80c1744). It computes the exact output size, rejects strings that cannot fit the `UNICODE_STRING` representation, checks both allocations/conversions, publishes the global strings only after complete construction, and honors explicit `InjectCmdLine` rules regardless of `--type=` arguments.

Validation: the complete `ClCompile` target for `SboxDll.vcxproj` passes in both `SbieRelease|Win32` and `SbieRelease|x64`, including `kernel.c`.

PR #5493, opened 2026-07-27
Add automatic snapshot capture on sandbox close

love-code-yeyixiao:
In `OptionsWindow.ui`, add a new checkbox `chkAutoSnapCapture` to automatically capture a snapshot when the last sandbox process terminates, and adjust the position and properties of related controls. Update the global settings to save this option.

In the `OnBoxClosed` function in `SandMan.cpp`, add a check for `AutoSnapCapture` to ensure that a snapshot is automatically captured and logged when the sandbox closes.

In `OptionsGeneral.cpp`, update the `LoadGeneral` and `SaveGeneral` functions to load and save the `AutoSnapCapture` option settings.


DavidXanatos:
please resove the conflict with the latest code base

offhub:
We can migrate this as `AutoDeleteSnapshotTarget=New` so that it works compatibly with the other two modes.

offhub:
@love-code-yeyixiao When `AutoSnapCapture` and `AutoDelete` are enabled together, `AutoSnapCapture` moves the files into a snapshot before auto-recovery runs. As a result, auto-recovery finds no files and does not display the recovery prompt. No files are lost, but this may confuse users who expect the recovery prompt to appear.

love-code-yeyixiao:
it now works well on my Win11 VM.

love-code-yeyixiao:
it now works well on my Win11 VM.

Kizuno18:
I found two failure-path problems in `OnBoxClosed` when automatic snapshots and automatic deletion are combined.

`TakeSnapshot` is synchronous and returns `SB_PROGRESS` errors such as `SB_SnapMkDirFail`, `SB_SnapCopyDatFail`, or a failed folder move, but the result was discarded and `DeleteBoxContent` ran immediately afterward. A failed snapshot could therefore still be followed by content deletion. Also, choosing the new plain **Delete Content** menu action sets `IsTakeSnapshot()` to false, but `bAutoSnap || TakeSnapshot` forced a snapshot anyway. Canceling the recovery dialog had the opposite inconsistency: it aborted deletion but still created a snapshot despite the comment saying capture happens immediately before deletion.

The fix checks and reports the snapshot result, aborts deletion on failure, preserves automatic capture when no recovery dialog is needed, and lets the dialog's primary/dropdown choice decide whether a snapshot is taken when the dialog is shown. I verified the cancel, empty-dialog, plain-delete, capture-success, and capture-failure branches with an actual-source regression harness; the PR UI also parses successfully with Qt 6.8.3 `uic`.

https://github.com/Kizuno18/sbxie/compare/f409ccad83c20f7087f85eba14d5c4c2c8d18a1b...461f69a46ef254ed966096fa9c19bed0fd5604d4

PR #5510, opened 2026-08-04
Add Junction redirect

love-code-yeyixiao:
Allow the user to configure virtual folder redirection only works to a certain sandbox.

Use JunctionPath=srcDir,dstDir to set a mapping
Use JunctionBlockRawAccess=y to enable raw access blocking(disable by default)

### DIFF
--- a/Sandboxie/core/dll/dll.h
+++ b/Sandboxie/core/dll/dll.h
@@ -185,6 +185,7 @@ typedef struct _THREAD_DATA {
     //
 
     ULONG file_dont_strip_write_access;
+    ULONG file_dont_block_junction_raw_access;
 
     //
     // misc modules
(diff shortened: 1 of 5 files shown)


love-code-yeyixiao:
I am working on resolving some known issues so it is not recommended to ask for Copilot reviewing before I cancel the draft status.

Kizuno18:
I found a concrete allocation-failure bug in `File_InitJunctions`: the new entry is counted before its strings are allocated, and the failure path resets `File_JunctionCount` to `index - 1`. Since `index` also includes skipped malformed `JunctionPath` lines, this can make later mapping scans and cleanup walk past initialized entries. The appended slot also has uninitialized pointer fields at that point.

I prepared a narrow fix in [409dda74](https://github.com/Kizuno18/sbxie/commit/409dda74ce61dc77da8bb3d02a0c18fda96d6518). It zero-initializes the slot, keeps the old count until initialization completes, and frees a partial entry on failure. The same patch also prevents mapping through a missing corresponding DOS/NT path form, checks the two temporary mapping allocations before use, and leaves room for `File_GetName` to restore a trailing root separator.

Validation: the complete `ClCompile` target for `SboxDll.vcxproj` passes in `SbieRelease|Win32` and `SbieRelease|x64`. Full solution builds compile `file.c` (including `file_junction.c`) on both platforms and then stop at resource compilation because this machine lacks the MFC `afxres.h` header.

PR #5547, opened 2026-08-21
Add zh_CN term check workflow

love-code-yeyixiao:
This PR aims to avoid deprecated translations from appearing in the future PR anymore.
I have confirmed it to run normally on my own repository.

### DIFF
--- a/.github/workflows/term.yml
+++ b/.github/workflows/term.yml
@@ -1,0 +1,115 @@
+name: Check deprecated Simplified Chinese terms
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, ready_for_review]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  check_pr:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check added Simplified Chinese translations
+        id: check
+        uses: actions/github-script@v9
+        with:
+          script: |
+            const simplifiedChineseFiles = new Set([
+              'Sandboxie/msgs/Text-SimpChinese-2052.txt',
+              'SandboxiePlus/SandMan/sandman_zh_CN.ts',
+              'SandboxiePlus/SandMan/Troubleshooting/lang_zh_CN.json',
+              'Installer/isl/ChineseSimplified.isl',
+              'CONTRIBUTING_zh_CN.md',
+              'CHANGELOG_zh_CN.md'
+            ]);
+
+            let violations = [];
+
+            try {
+              const changedFiles = await github.paginate(
+                github.rest.pulls.listFiles,
+                {
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  pull_number: context.issue.number,
+                  per_page: 100,
+                },
+              );
+
+              if (changedFiles.length < context.payload.pull_request.changed_files) {
+                core.setOutput('violations', '[]');
+                core.setFailed(
+                  `Unable to inspect all ${context.payload.pull_request.changed_files} changed files`,
+                );
+                return;
+              }
+
+              const targetFiles = changedFiles.filter((file) =>
+                simplifiedChineseFiles.has(file.filename),
+              );
+
+              const filesWithoutPatch = targetFiles
+                .filter((file) => file.patch === undefined)
+                .map((file) => file.filename);
+
+              if (filesWithoutPatch.length > 0) {
+                core.setOutput('violations', '[]');
+                core.setFailed(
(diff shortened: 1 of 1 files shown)


love-code-yeyixiao:
When the poll result in #5546 comes out I'll remove the final translation from the blacklist.

It works like:
image
image

love-code-yeyixiao:
Verified

love-code-yeyixiao:
Well would it be better to:

- specify a limited permission on the workflow

or
- use Github script as the core unit instead
?

.github/workflows/term.yml
```diff
@@ -0,0 +1,38 @@
+name: "Check Is PR included deprecated terms"
```
isaak654:
```suggestion
name: "Check if PR included deprecated terms"
```

.github/workflows/term.yml
```diff
@@ -0,0 +1,38 @@
...
+on: 
+  pull_request_target:
+    types: [opened, synchronize, ready_for_review]
+permissions:
+    pull-requests: write
+jobs:
+  check_pr:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check Is PR included deprecated terms
```
isaak654:
```suggestion
      - name: Check if PR included deprecated terms
```

.github/workflows/term.yml
```diff
@@ -0,0 +1,38 @@
...
+        id: check
+        uses: JJ/github-pr-contains-action@releases/v14.1.2
+        with:
+          github-token: ${{ github.token }}
+          diffDoesNotContain: "沙箱|沙盘"
+          waivedUsers: "dependabot[bot]"
+
+      - name: Post comment on failure
+        if: failure() && steps.check.outcome == 'failure'
+        uses: actions/github-script@v7
```
isaak654:
```suggestion
        uses: actions/github-script@v9
```

.github/workflows/term.yml
```diff
@@ -0,0 +1,38 @@
...
+              repo: context.repo.repo,
+              body: `🚫 **Failed to pass the term check**
+
+              Your pull request diff was checked to contain the following terms:
+
+              "沙箱""沙盘"
+
+              Unless the PR author explicitly confirms that this change is necessary, or this pull request is unrelated to localization changes, maintainers should not merge it. 
+              For the discussion on the Chinese translation of 'Sandbox', please refer to #5546.
+             `
```
isaak654:
```suggestion
              `
```

isaak654 (commented):
`箱|盘` is too broad and can block unrelated Chinese text. Please match the complete terms `沙箱` and `沙盘` instead, and limit the check to the intended zh-CN localization files.

love-code-yeyixiao:
I'm not sure about modifying a single character within a line. In the diff view of a GitHub Action, does the change show as the entire line or just that one character? If it's the latter, then if the PR author makes a modification based on the current wording, it won't trigger the detection. So maybe we still need to test it.

love-code-yeyixiao:
I have confirmed the comparion goes with diff by line, not by character.
So leave 沙 would be OK.

isaak654 (commented):
> [!NOTE]
> This is an untested code sample, feel free to consider this as an alternative approach to remove the dependency on `JJ/github-pr-contains-action` and limit the check to the intended zh-CN localization files

<details> 
  <summary><del>Code sample</del></summary>
</details>

love-code-yeyixiao:
Tested again. Look https://github.com/love-code-yeyixiao/Sandboxie/pull/24

DavidXanatos:
can you somehow compact it into one diff so that i can easily see all changes together

isaak654:
> can you somehow compact it into one diff so that i can easily see all changes together

Done.

Kizuno18:
The workflow can silently pass an incomplete scan on very large PRs. GitHub's `List pull requests files` endpoint returns at most 3,000 files, while `github.paginate()` does not indicate that more files were omitted. A PR with 3,001+ changed files can therefore place a targeted zh-CN file outside the returned set and avoid checking it.

I pushed a fail-closed guard that compares the returned list with the event's `changed_files` count: https://github.com/Kizuno18/sbxie/commit/6d5bb34d13c084c857ade78eeb22e1afb4485ea5

GitHub documents the 3,000-file limit here: https://docs.github.com/en/rest/pulls/pulls#list-pull-requests-files

I executed the workflow's actual embedded JavaScript with mocked API responses covering a complete clean diff, a deprecated term, a missing patch, and a truncated 3,001-file response; all paths behaved as intended with the fix.

PR #5557, opened 2026-08-23
Refine Traditional Chinese installer translations

mihomoQ:
Refined selected Traditional Chinese installer translations to use clearer Taiwan UI and technical terminology.


mihomoQ:
Superseded by #5553, which now includes these Traditional Chinese installer translation updates.

PR #5568, opened 2026-08-29
CHANGE ARABIC UI FROM RTL TO LTR

azizr12:
CHANGE ARABIC UI FROM RightToLeft to Left To Right please, because simply millions of arabic users are used LTR UI. and when you choose arabic you simply expect only the text to be RTL as arabic why change  the UI too ?????????
its such a problem that make using the app  harder and pain to use

>
> Please provide commit messages and pull request descriptions in English language.
>
> Translators are encouraged to look at the localization notes and tips: https://github.com/sandboxie-plus/Sandboxie/discussions/1123

### DIFF
--- a/SandboxiePlus/SandMan/SandMan.cpp
+++ b/SandboxiePlus/SandMan/SandMan.cpp
@@ -5050,6 +5050,9 @@ void CSandMan::LoadLanguage()
 	LoadLanguage(m_Language, "sandman", 0);
 	LoadLanguage(m_Language, "qt", 1);
 
+	if (m_Language.startsWith("ar", Qt::CaseInsensitive))
+		qApp->setLayoutDirection(Qt::LeftToRight); // keep it this way 
+	
 	QTreeViewEx::m_ResetColumns = tr("Reset Columns");
 	CPanelView::m_CopyCell = tr("Copy Cell");
 	CPanelView::m_CopyRow = tr("Copy Row");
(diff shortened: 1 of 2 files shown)


DavidXanatos:
I was not aware that this is not correct, are otehr supported languages affected as well? i see this fix targets 'ar' only?

azizr12:
wow such fast reply almost scared me
im not sure if my change is for all "ar" as global is correct i might add arabic (morocco)
but that itself will need a whole ar_ma files too
all i can say is the only language that may be needed for a fix too is hebrew
but hebrew translation isn't here

azizr12:
i tried the build from github it shouldnt  affect any language only arabic so its simple fix that i wanted

azizr12:
i will keep testing there is still some layout UI THAT need to be forced to be UI LTR

azizr12:
after some testing this code change only effect sandman.exe there are other executable that need its own fix too to stop this RTL UI

isaak654:
Merging this change would contradict the implementation in #4957. Maybe a feature request "Force LTR UI" would be better?

azizr12:
i think there should be two options either ar-ltr or ar-rlt simply add another arabic setting that do that
like ar_ma is treated as ar-ltr 
and ar_sa is treated as  ar-rlt 
that should fix the problem instead of forcing arabic globally to be RLF

PR #5586, opened 2026-09-04
update sandman_zh_CN.ts

habatake:
very minored chinese patch


mihomoQ:
I have two suggestions:
- Show Future Settings: 显示未来版本设置 would be clearer than 显示未来设置, since this option controls whether settings introduced in later versions are shown.
- Arabic (Morocco) - LTR: 阿拉伯语（摩洛哥，从左到右） reads more naturally in Chinese.

habatake:
Thank you, and fixed now

PR #5594, opened 2026-09-07
sandman: persistent identity profiles for volume serials

Kizuno18:
first functional slice of per-sandbox identity profiles: a persistent, versioned profile of volume serials that you create and edit in sandman, bind to a box, and whose values are actually consumed by the one mechanism that supports them today, the `GetVolumeInformationByHandleW` hook via `HideDiskSerialNumber` + `DiskSerialNumber`. no driver, svc or dll change, based on `master` and independent of #5593 and of the volume serial cache branch (that one improves cache keying inside the hook, this one decides what values the hook is fed)

what you get

- `Helpers/IdentityProfile.{h,cpp}`: `CIdentityProfile` (version 1, id, name, revision, `HarddiskVolumeN` -> `XXXX-XXXX`), validation and normalization (`\Device\HarddiskVolume1` accepted, drive letters / `PhysicalDriveN` / duplicates rejected), `CIdentityProfileStore` (one json per profile under the sandman data dir, `QSaveFile` atomic write + read-back, list/load/save/remove/export/import), `CIdentityProfileBinding` (apply/unbind/read state against a small `CIdentityIniTarget` interface)
- `Windows/IdentityProfilesDialog.{h,cpp}`: list dialog (new, edit, clone, regenerate, import, export, delete) and the editor with an explicit **Regenerate Serials** button
- advanced > privacy: a profile combo next to *Hide Disk Serial Number* plus **Identity Profiles...**; apply/ok binds or unbinds through the normal `SaveAdvanced` path, errors surface via `SB_STATUS` like every other option
- `IdentityProfile` / `IdentityProfileRevision` box settings, documented in `SbieSettings.ini`; changelog line; `Docs/IdentityProfiles.md` and the 22-category `Docs/IdentityCoverage.md` matrix (only category 8 and the profile plumbing in 1/15/19/20 move in this slice, everything else is marked as not in this slice or preserved real)

the rules the binding enforces

- the profile is the source of the values; the box section is the applied copy; nothing here claims verified protection, the ui text says so
- apply and unbind refuse while the box has running processes (already running dlls keep their values), and binding refuses over manual `DiskSerialNumber` entries, remove those first
- regenerate is explicit, bumps the revision, and bound boxes show *stale* until re-applied while stopped; nothing is rewritten on its own. raw ini edits show as *diverged*, a deleted profile as *missing*
- clone = new id, same serials; import = always a new id, serials preserved, never overwrites; export goes through `QSaveFile`
- templates and the global section are never written, `Template=` references stay, and boxes without a profile keep exactly their current `HideDiskSerialNumber` behaviour (unbind leaves that checkbox as it was)
- no environment variable, seed or external process is an authority for the profile, and the store only ever touches its own directory

what it does not do, on purpose

- only the by-handle volume query: `GetVolumeInformationW/A`, `NtQueryVolumeInformationFile`, wmi and physical disk serials are not covered and not claimed
- this is the **volume** serial (`vol`), not the physical disk serial
- `HarddiskVolumeN` is a name windows can reassign, not a physical volume binding
- no runtime "protection active" indicator, no fail-closed launch, no service-side atomic ini transaction (profile file is atomic, the ini write is read-back verified only)
- no rollback of a half-applied ini write beyond reporting the failure; the binding state tells you what is there
- the profile file itself now has compare-and-swap on its revision and a per-profile lock (see the follow-up); the "no concurrent-writer protection" caveat is about the box ini section, not the profile store

tests, three kinds kept apart

- simulated: new `SandboxiePlus/tests/identity-profile` ctest target, 28 cases compiling the shipped helper and dialog sources unchanged against a `QTemporaryDir` store and a synthetic box target. covers persistence across store instances, atomic save failure with the old file intact, invalid input and corrupt json, multiple volumes, two profiles across three boxes, refusing running boxes and manual entries, stale/diverged/missing states, templates untouched, host untouched, and the dialog flows (new/edit/clone/regenerate/import/export/delete guard), plus `bind-deferred-commit` for writes that only become visible on commit. 28/28 locally with g++ 16.2 + qt 6.11 offscreen
- build, split by revision (all on my fork, none is an upstream check):
  - 23b54d07, runs https://github.com/Kizuno18/sbxie/actions/runs/34160114364 and https://github.com/Kizuno18/sbxie/actions/runs/34160253007: qt5 and qt6 27/27 both times; the windows job died in workflow plumbing before touching the code (cmake configure against the repo qt package, which ships no `Qt6EntryPointd.lib`, then a wrong `jom.exe` path). no code evidence from those two windows jobs
  - 23b54d07, run https://github.com/Kizuno18/sbxie/actions/runs/34160376589: windows msvc + repo qt6 ran all 27 cases, 27/27, but the real `SandMan.exe` link failed with `LNK2019` on `CIdentityProfileStore`/`CIdentityProfilesDialog` symbols: the new sources were in `SandMan.vcxproj` but not in `SandMan.pri`, and `qmake_plus.cmd` builds from the qmake project. real code gap, fixed in 06af52de
  - 06af52de, run https://github.com/Kizuno18/sbxie/actions/runs/34166990071: qt5 27/27, qt6 27/27, windows msvc 27/27, `QSbieAPI.dll` and `SandMan.exe` x64 build and link. compile-and-link proof for the `.ui`/`OptionsAdvanced` integration, nothing more
  - 6ef86f5f (current head), run https://github.com/Kizuno18/sbxie/actions/runs/34169383236: qt5 28/28, qt6 28/28, windows msvc 28/28, `QSbieAPI.dll` and `SandMan.exe` x64 build, sha256 and blob hashes in `identity-manifest.txt`, which match this branch
- runtime, done by hand on a lab vm with the real driver and the real sandman ui against 6ef86f5f; the probe script and the exact steps are in the branch (`tests/identity-profile/volume_probe.ps1`, `tests/identity-profile/RUNTIME.md`), the anonymised result table too. 664ad0fb on top is that documentation only, no code change after 6ef86f5f
  - lab conditions: windows 10 x64 vm, official sandboxie-plus 1.18.4 installer, then `SandMan.exe`/`QSbieAPI.dll` swapped for the fork ci artifacts of 6ef86f5f (hashes checked against the run manifest). the official driver refuses an unsigned manager as session agent (`STATUS_INVALID_SIGNATURE`), so the vm ran in test-signing mode where `MyIsCallerSigned` short-circuits: that is a lab condition for running a development manager, not a requirement of the feature and not a statement about production security. two volumes, C: and a fixed vhd D: (`HarddiskVolume3`/`HarddiskVolume4` there), two empty boxes, one host control process outside any box
  - the same ui flow on 06af52de hit a real bug: "identity profile not applied: configuration read back does not match the profile" although the ini was correct once `SaveConfig` committed (`SaveConfig` disables refresh-on-change, sbiesvc publishes batched writes on commit only, the read-back ran before). 6ef86f5f fixes it with a `Flush()` hook on the target (sandman adapter = `CSbieIni::CommitIniChanges()`) before every read-back, unbind verifies too, `bind-deferred-commit` models it in the suite
  - with 6ef86f5f: profile a bound to BoxA, profile b bound to BoxB through the dialog, no error dialog; `Sandboxie.ini` shows per box `HideDiskSerialNumber=y`, `IdentityProfile=<id>`, `IdentityProfileRevision=1` and the two `DiskSerialNumber` rows; two json files with revision 1 in the sandman data dir
  - observed by `GetVolumeInformationByHandleW` inside the boxes: BoxA `A1A1-0001`/`A1A1-0002`, BoxB `B2B2-0001`/`B2B2-0002`, stable over fresh processes and over a reboot. per volume: C: was checked in every run; D: only counts where the vhd was attached, so after the reboot the same vhd was reattached and both boxes plus the host were queried again without recreating or re-applying anything, and both volumes kept their values in both boxes. while BoxB was briefly bound to profile a (keyboard slip in the combo) it showed the A1A1 values, i.e. a shared profile behaves as designed. the host control kept its real serials on every run, before and after; those values are not published
  - still not claimed: physical disk serial, wmi, native `NtQueryVolumeInformationFile`. `GetVolumeInformationW` happened to return the same substituted value inside the boxes on this windows build, recorded as an observation only. no runtime "protection active" indicator, no fail-closed launch, no service-side transaction

Follow-up: profile saves now take the same per-profile lock as removal and check the stored revision before replacing the file. A stale editor cannot overwrite newer serials using the same revision or recreate a profile that another instance deleted. Two new regressions failed on 664ad0fb (verified: `store-stale-writer`, `store-locked-writer` fail there) and pass with this fix. Linux/GCC 16.2.1/Qt 6.11.2: `cmake -S SandboxiePlus/tests/identity-profile -B build/identity-profile -DCMAKE_BUILD_TYPE=Release`, `cmake --build build/identity-profile -j4`, and `ctest --test-dir build/identity-profile --output-on-failure` all exit 0, 30/30 cases, and the fork run https://github.com/Kizuno18/sbxie/actions/runs/34241271777 on `dbc47de4` (current head `f498bfe9` is the same tree plus a CHANGELOG-only commit) is green on qt5, qt6 and windows msvc + the real `SandMan.exe`/`QSbieAPI.dll` x64 build. This follow-up has not been tested against Windows file locking or the full manager; the earlier Windows/runtime evidence above applies only to its recorded revisions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [1.18.4 / 5.73.4] - 2026-09-06
 
 ### Added
-- added per-sandbox identity profiles to SandMan (Advanced > Privacy): persistent, versioned volume serial profiles that are bound to a sandbox and written as `DiskSerialNumber`/`HideDiskSerialNumber`; covers the existing `GetVolumeInformationByHandleW` hook only
+- added per-sandbox identity profiles to SandMan (Advanced > Privacy): persistent, versioned volume serial profiles that are bound to a sandbox and written as `DiskSerialNumber`/`HideDiskSerialNumber`; covers the existing `GetVolumeInformationByHandleW` hook only. Saves take a per-profile lock and check the stored revision, so a stale editor cannot overwrite newer serials or recreate a removed profile
 - added low-noise `SbieTrace` logging for `Shell_NotifyIconW` calls, including the message, icon identity (`NIF_GUID`/GUID or `HWND`/`uID`), and effective direct/proxy route
 - added the `Show Future Settings` editor setting (enabled by default) to include future-version settings in SandMan INI completion candidates
 
 ### Fixed
-- fixed stale identity profile edits overwriting newer serials and saves racing profile removal
 - fixed stale INI completion-popup candidate tooltips during editing and limited `Template`/`TemplateReject` tooltips to setting-name text
 - fixed SandMan File Panel "Create Shortcut" producing a shortcut without a working directory, so the sandboxed program inherited SandMan's current directory and applications that open their data files by relative path failed to find them [#5542](https://github.com/sandboxie-plus/Sandboxie/issues/5542)
 - fixed wrong return type ein sbiesvc
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [1.18.4 / 5.73.4] - 2026-09-06
 
 ### Added
+- added per-sandbox identity profiles to SandMan (Advanced > Privacy): persistent, versioned volume serial profiles that are bound to a sandbox and written as `DiskSerialNumber`/`HideDiskSerialNumber`; covers the existing `GetVolumeInformationByHandleW` hook only
 - added low-noise `SbieTrace` logging for `Shell_NotifyIconW` calls, including the message, icon identity (`NIF_GUID`/GUID or `HWND`/`uID`), and effective direct/proxy route
 - added the `Show Future Settings` editor setting (enabled by default) to include future-version settings in SandMan INI completion candidates
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - added the `Show Future Settings` editor setting (enabled by default) to include future-version settings in SandMan INI completion candidates
 
 ### Fixed
+- fixed stale identity profile edits overwriting newer serials and saves racing profile removal
 - fixed stale INI completion-popup candidate tooltips during editing and limited `Template`/`TemplateReject` tooltips to setting-name text
 - fixed SandMan File Panel "Create Shortcut" producing a shortcut without a working directory, so the sandboxed program inherited SandMan's current directory and applications that open their data files by relative path failed to find them [#5542](https://github.com/sandboxie-plus/Sandboxie/issues/5542)
 - fixed wrong return type ein sbiesvc
--- a/Sandboxie/install/SbieSettings.ini
+++ b/Sandboxie/install/SbieSettings.ini
@@ -2590,6 +2590,32 @@ Syntax=[sn]=[bN]
 Description=When enabled, the box border overlay windows are excluded from\nscreen capture (using [code]WDA_EXCLUDEFROMCAPTURE[/code]). When not set,\nthe value of [code]CoverBoxedWindows[/code] is used as the default. Can be set globally or per-box.
 
 
+[IdentityProfile]
+AddedVersion=1.18.4
+RemovedVersion=
+ReAddedVersion=
+RenamedVersion=
+SupersededBy=
+Category=ay
+Context=
+Requirements=
+Syntax=[sn]=profile_id
+Description=Id of the SandMan identity profile bound to this sandbox.\nSandMan writes the profile's volume serials as DiskSerialNumber entries and sets HideDiskSerialNumber; the profile itself is stored by SandMan, not in this file.
+
+
+[IdentityProfileRevision]
+AddedVersion=1.18.4
+RemovedVersion=
+ReAddedVersion=
+RenamedVersion=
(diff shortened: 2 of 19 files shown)


mihomoQ:
There seems to be a conflict between the existing checkbox and profile binding when both change in the same save.

Start with an unbound sandbox that has `HideDiskSerialNumber=y`, uncheck “Hide Disk Serial Number”, select an identity profile, and apply.

`SaveAdvanced()` first queues the deletion of `HideDiskSerialNumber=y`. Profile binding then calls `SetText(..., "y")`, but `GetText()` still returns the old value, so `SetText()` skips the write. After `Flush()`, the profile binding and serials are saved, but the hide setting is missing, and read-back reports a mismatch.

I reproduced this with the PR’s binding code and a deferred test target that includes the equality check from `CSbieIni::SetText()`, not through the full GUI. The existing `SFakeBox::SetText()` always writes, so it misses this case.

Could the checkbox write be skipped when applying a profile, since the profile sets this option itself?

Kizuno18:
@mihomoQ Fixed in 57f4f66f. The checkbox write is skipped when a profile will be applied; manual/template saves and unchanged missing-profile bindings keep the checkbox path.

The deferred fake now includes the SetText equality check. Added regressions for the reported save and the missing-profile case; all 32 harness tests pass. Validation used the deferred target, not the full Windows GUI.

PR #5596, opened 2026-09-08
Added error handling and UI refresh

Code-LSQ:
Affects Sandboxie-Plus only

1. Added error handling during extension installation to prevent silent failures caused by network issues (especially for users in China) that would previously show as successful installation.

2. Refresh UI state after removing an extension; the previous behavior kept showing it as still installed after removal.

Additionally, I noticed that the latest version of SbieHide is now 2.0.1, while Sandboxie detects it as 1.0. I originally planned to modify the extension detection logic, but I found that the Addon interacts with sandboxie-plus.com/update.php to fetch version information, so I did not change that part.

Is it possible to modify the detection logic for the latest extension version? If this is rejected for compatibility or security reasons, would it be possible to add support for the latest SbieHide version? https://github.com/VeroFess/SbieHide/releases/tag/v2.0.1 — it seems they changed the file naming and added new DLL. I manually downloaded it and specified InjectDll manually, and it works normally.

Also, the check for ImDisk in the extensions seems to be behind the latest version as well?

### DIFF
--- a/SandboxiePlus/SandMan/AddonManager.cpp
+++ b/SandboxiePlus/SandMan/AddonManager.cpp
@@ -7,6 +7,7 @@
 #include <QUrlQuery>
 #include <QJsonDocument>
 #include <QJsonObject>
+#include <QSettings>
 #include "../QSbieAPI/Sandboxie/SbieTemplates.h"
 #include <QtConcurrent>
 #include "../../SandboxieTools/UpdUtil/UpdUtil.h"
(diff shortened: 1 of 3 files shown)


Kizuno18:
The functional changes at the current head (`e064ecf41305bd3ee4e5c4c911e1b0d1c14f2c74`) look useful: the typed `CSbieProgress::Finished` connections compile and the updater now returns `ERROR_GET` when add-on metadata cannot be downloaded.

There is one small merge-hygiene issue: `git diff --check` reports trailing whitespace in the newly added `QMessageBox::warning` call at `AddonManager.cpp:48`. I pushed the one-character cleanup here:

- commit: https://github.com/Kizuno18/sbxie/commit/ef3561ed
- comparison: https://github.com/sandboxie-plus/Sandboxie/compare/fix...Kizuno18:sbxie:codex/help-pr-5596

Validation: with Qt 6.8.3/MSVC, `AddonManager.cpp` and `SettingsWindow.cpp` both compile successfully, including the typed signal/slot expressions. The x64 `UpdUtil` `ClCompile` target also succeeds. After the cleanup, `git diff --check` is clean.

Code-LSQ:
Thank you for your correction. I have made the changes as suggested.

mihomoQ:
One edge case here: the new check also returns early for `UpdUtil.exe modify remove:<id>` without `/step` if the metadata download fails.

Previously, that path could continue to `LoadAddon()` and remove the add-on using its local JSON record. The GUI passes `/step:apply`, so GUI removal is unaffected.

Could the check preserve this local removal path while still reporting download failures when metadata is actually needed? This is based on reading the code; I haven't tested an offline uninstall.

Code-LSQ:
In 62abac138b5a386ed38584fb1e8a9c7db491bb48, I fixed a bug that only manifests in some cases: the UI did not refresh after addon deletion.


@mihomoQ You are right. I added handling for this case in 62abac138b5a386ed38584fb1e8a9c7db491bb48.

PR #5597, opened 2026-09-08
sandman: add redacted stack trace exports

Kizuno18:
sandman gains a redacted stack trace export in the trace-view stack panel: the local capture details stay in the dialog, `Copy` and `Save` produce a share view built from an allow-list. it never renames the bridge to a real vendor, never hides the sandboxie modules from anything else, and never touches the underlying capture; it only anonymises the diagnostic you decide to share

why the split matters

- **local details**: full symbol strings and addresses as they come from the resolver, visible only inside this dialog. useful for local debugging
- **share preview**: the exact bytes that copy or save will write, built from the transformer. what you see before you decide is what leaves the dialog
- the two are computed from one immutable capture: a live diagnostic refresh cannot swap the share content out from under a copy/save the user just approved

what changes

- new `Helpers/RedactedStackExport.{h,cpp}`: a pure transformer with an immutable `SStackCapture` input (address, symbol string, opaque module key) and an `SRedactedStack` output that only carries allow-list fields (`index`, opaque per-report `moduleId`, `moduleResolved`/`symbolResolved` booleans, a canonical `line`). writer is `QSaveFile` with `directWriteFallback` off, so a failed commit leaves the previous file intact and no temporary sibling survives
- new `Windows/StackExportDialog.{h,cpp}`: the two-pane dialog. text/json format switch; the same transformed bytes go to both preview and file
- `Views/StackView`: context-menu action **Export Redacted...** that builds a capture from the currently shown stack and pops the dialog. no live-diagnostic path is duplicated
- one changelog line

what it does not do, on purpose

- no re-attribution: markers say they are markers (`[module-01]`, `[symbol redacted]`, `[unresolved]`). never a real product name
- no hiding of sandboxie from anything: the sandboxie modules keep appearing in the diagnostic and in the local details, only the share view omits their names
- paths, absolute addresses and offsets, module and symbol names, capture owner, box name, arguments and free-form error strings never leave the transformer
- opaque module ids are valid within one report only; correlation between reports is not prevented and the note in the json says so
- no dumps, no screenshots, no upload, no changes to third-party dlls, no changes to the existing #5593/#5594 branches

tests, three kinds kept apart

- simulated: new `SandboxiePlus/tests/redacted-stack-export` ctest target, 16 cases against the shipped transformer with a `QTemporaryDir` store and a synthetic capture, driven by an allow-list of private tokens that must never appear in the text, the json, the on-disk file or the error string. covers allow-list keys, opaque-id semantics (order-of-appearance per report, distinct modules stay distinct, same key inside a report shares its id), same-name-distinct-modules vs same-name-shared, frame order, capture-state (`complete`/`partial`), unresolved frames marked (never guessed), unicode/punctuation, text/json round-trip, atomic save happy path, open failure, commit failure, existing-file replacement, input untouched, two exports starting the id space at 1 each
- build: fork run https://github.com/Kizuno18/sbxie/actions/runs/34245372447 on `d1429ad3` — qt5 and qt6 build and run 16/16, msvc + repo qt6 build and run 16/16, then the real `QSbieAPI.dll` and `SandMan.exe` x64 build and link, sha256 in `redacted-manifest.txt` next to the commit and the blob hashes of the new sources (the "not an upstream check" caveat applies)
- runtime: not part of this pr. real symbols come from dbghelp on a windows host with sandboxie installed; the runtime check is manual, driving the "Export Redacted..." action on a live capture and comparing the share view with what this transformer produces

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [1.18.4 / 5.73.4] - 2026-09-06
 
 ### Added
+- added a redacted stack-trace export in the SandMan trace view: the local stack details stay in the dialog, `Copy` and `Save` produce a share view built from an allow-list (opaque per-report module ids, no paths, no absolute addresses/offsets, no module or symbol names) and the atomic writer never falls back to a direct write
 - added low-noise `SbieTrace` logging for `Shell_NotifyIconW` calls, including the message, icon identity (`NIF_GUID`/GUID or `HWND`/`uID`), and effective direct/proxy route
 - added the `Show Future Settings` editor setting (enabled by default) to include future-version settings in SandMan INI completion candidates
 
--- a/SandboxiePlus/SandMan/Helpers/RedactedStackExport.cpp
+++ b/SandboxiePlus/SandMan/Helpers/RedactedStackExport.cpp
@@ -1,0 +1,158 @@
+#include "RedactedStackExport.h"
+
+#include <QJsonArray>
+#include <QJsonDocument>
+#include <QMap>
+#include <QSaveFile>
+
+//---------------------------------------------------------------------------
+// SRedactedStack
+//---------------------------------------------------------------------------
+
+QJsonObject SRedactedStack::toJson() const
+{
+	QJsonObject json;
+	json["format"] = "sandboxie-stack-export";
+	json["version"] = version;
+	json["captureState"] = captureState;
+	json["frameCount"] = frameCount;
+	json["moduleCount"] = moduleCount;
+	QJsonArray array;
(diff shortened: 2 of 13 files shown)


DavidXanatos:
I don’t think I can approve this in its current form.

For debugging, the DLL/module names and resolved symbols or exports are essential parts of a stack trace. Redacting those leaves mostly the call depth and module grouping, which makes the resulting trace of very limited diagnostic value.

I think the privacy filtering should instead focus on genuinely sensitive information such as full filesystem paths, usernames, sandbox names, absolute addresses, and similar user-specific data.

Module basenames like ntdll.dll, chrome.dll, SbieDll.dll, etc. should remain visible, as should resolved symbol/export names. Otherwise the exported trace loses most of the information needed to understand what actually happened.

